PR #12, opened 2016-08-16
Fix shbang parsing edge cases.

### LINKED ISSUE #11: yath won't run if perl is in #! twice
```
$ cat ~/tmp/test.t
#!/perl/blah/perl

use Test::More;

diag "Perl is $^X";
pass;

done_testing;
```

```
$ yath ~/tmp/test.t
[LAUNCH]  _  /Users/schwern/tmp/test.t
[STDERR]  +  Can't open perl script "/blah/perl": No such file or directory
[ PLAN ]  =  No plan was ever set.
[FAILED]  =  /Users/schwern/tmp/test.t

=== FAILURE SUMMARY ===
 * /Users/schwern/tmp/test.t
```

I noticed this while trying yath with Regexp::Common. It has `#!/opt/perl/bin/perl`. I'm wondering why yath is even trying to open the #! perl.

schwern:
This borrows from TAP::Parser::SourceHandler::Perl

Fixes #11

### DIFF
--- a/lib/Test2/Harness/Runner.pm
+++ b/lib/Test2/Harness/Runner.pm
@@ -111,6 +111,31 @@ sub start {
     return $self->via_do(@_);
 }
 
+sub _parse_shbang {
+    my $self = shift;
+    my $line = shift;
+
+    return {} if !defined $line;
+    
+    my %shbang;
+
+    my $shbang_re = qr{
+        ^
+          \#!.*\bperl.*?        # the perl path
+          (?: \s (-.+) )?       # the switches, maybe
+          \s*
+        $
+    }xi;
+    
+    if ( $line =~ $shbang_re ) {
+        my @switches = grep { m/\S/ } split /\s+/, $1 if defined $1;
+        $shbang{switches} = \@switches;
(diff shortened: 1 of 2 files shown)


lib/Test2/Harness/Runner.pm
```diff
@@ -124,19 +147,16 @@ sub header {
     );
 
     open(my $fh, '<', $file) or die "Could not open file $file: $!";
-    my $ln = 0;
-    while (my $line = <$fh>) {
-        $ln++;
+
+    my $shbang = $self->_parse_shbang(<$fh>);
```
exodist: This assumes the first line of the test is a #!, what if the test has no #! line and instead starts with a directive tot he harness? This will miss it.

exodist: This is something I thought about, and intended to test for, but never got around to it.

schwern: Good point, I'll fix that.

schwern:
Fixed!

exodist:
+1

PR #13, opened 2016-09-01
Pass the Job object to job listeners rather than just the id

autarch:


exodist:
looks good ot me, feel free to merge it once the travis testing completes.

PR #15, opened 2016-09-05
Add support for passing args to tests after a double dash (--)

autarch:


autarch:
I'd like to have more tests for this, but I'm really not sure how to implement them. @exodist, any thoughts?

I've been testing this by hand, and it does work, but I feel like it's the kind of feature that might be easy to break in a refactoring without noticing.

schwern:
I have some questions.

What's the rationale for `::` meaning "pass the rest of the arguments to the tests"? How will the user know or remember that association? Could it be something more descriptive and conventional like `--args-for-tests='--test-arg --more'`?

How is this better than using environment variables? Are users likely to use it instead of environment variables?

exodist:
I was assuming there was a similar feature in Test::Harness, but instead of assuming, I will just ask, is there something out there that already does this?

exodist:
http://stackoverflow.com/questions/13632377/prove-command-not-passing-arguments-after-arisdottle-to

Could not find documentation on how to do it, but did find docs that it is in fact a real thing in Test::Harness that people may expect to have in Test2::Harness

autarch:
Yes, Test::Harness does support this exact feature. It's documented at https://metacpan.org/pod/distribution/Test-Harness/bin/prove#Arguments-to-Tests

In fact, I more or less copied those docs for my patch.

As for why not use env vars - it's kind of a pain, especially if you want to run the same command repeatedly with slightly different args, since you set those at the beginning of the command line.

The `--test-args="..."` is just a non-starter IMO. It adds all sorts of very confusing questions about quoting in the shell, escaping, etc.

autarch:
Rebased to fix the merge conflict.

schwern:
[This has been around since 2007](https://github.com/Perl-Toolchain-Gang/Test-Harness/commit/463e71dcf2cbf56a40426b5fd76db87a55ee621b) and it's the first either Chad or I have heard of it. It's not a convention of any other program I know of.

Environment variables solve the same problem. They work without special harness support. They're an existing convention we don't have to add docs, code, tests, and educational material for.

> As for why not use env vars - it's kind of a pain, especially if you want to run the same command repeatedly with slightly different args, since you set those at the beginning of the command line.

`ctrl-a` will jump you to the beginning of the line in most shells, just as `ctrl-e` will jump to the end.

I don't think it provides anything that env variables don't. I don't think users will know about it. I think it will be largely unused. Sorry, I don't see the value.

autarch:
I think the main value is that this is what `prove` does (even if it's weird, which it is). If you're using `Test::Class::Moose` this is a feature you're likely to use a lot because of how TCM works. You have one `.t` file but many test classes, so you need a way to pass just a single class to the test runner invoked by the `.t` file. The use of the arisdottle (::) is documented in https://metacpan.org/pod/Test::Class::Moose::Runner as the way to achieve this.

Yes, I know how to use `ctrl-a` and such, but it's just a touch more tedious than `↑`, `ctrl-w`, `New::Test::To::Run`.

That said, I could always just write a little wrapper around yath to set the env var for me anyway, of course.

schwern:
If TCM is the only use case, could we patch TCM to take environment variables?

autarch:
The args aren't processed by TCM directly. You write code that processes them. It's just a lot simpler to use `Getopt::Long` (or its equivalent) than to have to write your own code to parse a whole bunch of stuff from an env var. It also occurs to me that the env var has the exact same quoting/escaping problems as `--test-args=""`.

schwern:
@autarch Are you assuming that you'll pass in a bunch of things as a single env variable? Instead, use one env variable for each thing. No extra code necessary.

    my $Class = $ENV{THE_CLASS_FOR_THIS_TEST};
    my $More = $ENV{DO_MORE_STUFF} ? 1 : 0;

autarch:
What if I want to specify three classes or something like that? I think you'd still end up having to parse it a bit and come up with some sort of convention for separate the names (spaces, colon, whatever). It also seems kind of painful, especially when we already have a precedent in `prove` that people are used to it (for values of "people" that include the people who've found a need for the feature).

What exactly is so horrible about this particular feature that you don't want to include it? It's not a lot of code, it's simple to document, and it's proven itself useful in prior art.

schwern:
> What if I want to specify three classes or something like that?

This edge case is moving ever closer to the edge. Split on commas?

> especially when we already have a precedent in prove that people are used to it

One is not a precedent. One is someone tried something out. I'm not convinced that people know about and use this feature. I could be wrong.

> What exactly is so horrible about this particular feature that you don't want to include it? It's not a lot of code, it's simple to document, and it's proven itself useful in prior art.

In the end it is a fairly minor thing and it is ready to go, which is why I'm not closing it if someone else wants to merge it. So I'll wrap up my position:

* It doesn't do anything that environment variables don't already do.
* Environment variables are well understood, this is not.
* Use of this feature (ie. `@ARGV`) requires special harness support, environment variables do not.
* `::` is not a command line idiom.
* `::` is not descriptive about what it does, it's more cryptic line noise for the reader to puzzle out.

The showstopper for me is that last one, it's bad UI. Some poor slob will read `yath :: --foo --bar=23` in some CI configuration and wonder what is going on. It's line noise to support an edge case that already has a good and well understood solution. I see this as another quirky, obscure, Perl-only feature that adds maint load and makes Perl projects harder to understand. I haven't heard a good argument otherwise. "prove does it" doesn't count.

Maybe there's a more readable way to do the same thing?

If it turns out this is an important compatibility feature to get people to switch from prove, then ok.

autarch:
So I'm working on some things related to Test::Class::Moose that got me thinking about this again. Here's a very concrete example. Let's say I have a file named `tcm.t`. By default, running that file will run all of your test classes in the normal way. But we want to be able to somehow pass information to that script to set a variety of options. Here's how this might look from the command line using `prove`:

    prove -lv t/tcm.t :: --classes TestFor::Foo --classes TestFor::Bar --exclude-tags 'Slow Methods' --timing-data-file ./path/to/timing.json

This is quite simple to read, and importantly, it's simple to parse inside tcm.t because we can just use `Getopt::Long` (or `MooseX::Getopt` really, but whatever).

The equivalent env var version is this:

    TCM_CLASSES=TestFor::Foo,TestFor::Bar TCM_EXCLUDE_TAGS="Slow Methods" TCM_TIMING_DATA_FILE=./path/to/timing.json yath -lv t/tcm.t

I guess that works, but it's pretty gross, and there are many possible arguments that can be set. Also, I still can't pass `--help` to `t/tcm.t` to get it to tell me what it accepts.

I really wish I knew the history of why this was added to prove. Looking back in the code, it started out as `--` and then was changed to `::` in https://github.com/Perl-Toolchain-Gang/Test-Harness/commit/f17df7bdba6b376aa7c93106da7370701380d855

Unfortunately, the commit doesn't explain why he switched it. I think `--` is more standard, and I'd be happy to use that instead, except for the fact that `prove` doesn't.

exodist:
I think it is silly to try to carry forward this hack. Here is my reasoning:

The only reason this is even a thing is because Test::Class and similar need a .t file to act as a launcher between the harness and the test classes. Test::Harness (prove) does not understand this at all, so you need to use the hack to pass arguments into what is essentially the ONLY test file (the .t that launches the classes).

So I say we need to step back. I think we should make Test::Harness aware of this pattern. Specifically I think we should have the concept of a 'launcher'. The launcher may be a class name passed in, or a script file to run. For current Test::Class and similar we can say something like `--launcher foo.t` and then it will be smart enough to pass any arguments after that into the launcher. `--launcher-class Test::Class` should be smart enough to load 'Test2::Harness::Launcher::Test::Class'  which will basically take over for the .t file you have to have as a prove compatibility shim.

Test2::Harness is a good place to do things right, I do not want to simply have it do what 'prove' does without looking at the 'why' first.

autarch:
What's the difference between `yath -lv --launcher t/tcm.t --classes ...` and `yath -lv t/tcm.t -- --classes ...`? It's actually _less_ standard that way.

As I pointed out, the use of `--` to mean "stop processing CLI args here" is pretty common (see `find` and `grep`, among others). I don't know why prove uses `::` instead, but the concept is not new.

exodist:
Let me be more clear. I am fine with making `--` work, but I am not ok with `::`. If we figure out why they used `::` and determine it is applicable to us then I want to step back and discuss instead of just doing `::`.

Separately, I want the pattern where a single .t file launches other test files to be less necessary, it would be nice to have the harness solve the launch problem where possible.

autarch:
Ping ... I'm pretty sure the AppVeyor failure is not related to this change.

exodist:
-- and :: are both supported in current Test2::Harness.

PR #18, opened 2016-09-14
Added a new DataDumper renderer

autarch:


lib/Test2/Harness/Fact.pm
```diff
@@ -242,7 +230,7 @@ True if the fact results in a failure. (Example: 'not ok')
 
 True if the fact adds to the test count.
 
-=item $plan_ref = $f->sets_plan()
```
exodist: What is the point of removing the parens from the docs here?

lib/Test2/Harness/Renderer/JSONDump.pm
```diff
@@ -0,0 +1,122 @@
+package Test2::Harness::Renderer::JSONDump;
+use strict;
+use warnings;
+
+our $VERSION = '0.000012';
+
+use File::Basename qw/basename dirname/;
```
exodist: Are any of these new deps non-core? Just curious, not a blocker or anything.

lib/Test2/Harness/Fact.pm
autarch: Just trying to keep the docs consistent.

lib/Test2/Harness/Renderer/JSONDump.pm
autarch: Nope, all core. If non-core was okay it'd be nice to use Path::Tiny or something, but this works well enough.

autarch:
This needs some more testing against real test suites. Please don't merge yet.

PR #19, opened 2016-09-15
Add a renderer which uses Data::Dumper

autarch:


exodist:
Do you still want this? if so it needs an update for fact->event change. IF this no longer interests you please close it.

autarch:
I don't think I really need this any more. If it comes up again I can always rework it.

PR #20, opened 2016-12-05
Convert all uses of facts to events

autarch:
These changes rely on changes currently in a branch of the test-more repo to
make subtests which are marked todo set effective_pass to true on their
contained events.

This is a rather huge change that does the following:

* All uses of Test2::Harness::Fact have been replaced by Test2::Event objects
  * Wherever possible, we use events already part of the test-more distro.
  * This also adds many new events specific to the harness, for example
    ProcessStart and ProcessFinish events.
  * The T2Harness output and EventStream parser replay events _exactly_ from
    the .t files, while the TAP parse has to reconstruct events by parsing the
    TAP, meaning that they're not going to be exactly the same. In particular,
    the TAP parser events don't have a trace.
* Lots of simplification of the guts of the Result and Job classes. In
  particular, the Result class was replicating functionality already provided
  by using Subtest events instead of facts.
* Removed checking the test numbering in the TAP parser. This could be added
  back, but I'm not sure how much effort this is worth. Parsing TAP is fragile
  and hard to perfect. We should encourage people to move to code that can use
  the new T2Harness output if they want to use this distro.
* Lots of changes to the tests, including ripping out a fair bit of mocking
  and using actual event objects instead. The mocking didn't really achieve
  much because the thing being mocked (a Fact, and now an Event) is really
  easy to construct.
* The Test2::Harness::Parser is now explicitly an abstract base
  class/factory. You can instantiate one but it no longer provides parse_*
  methods.


lib/Test2/Event/Encoding.pm
```diff
@@ -0,0 +1,17 @@
+package Test2::Event::Encoding;
```
exodist: This event belongs in Test2 itself

lib/Test2/Event/ParseError.pm
```diff
@@ -0,0 +1,20 @@
+package Test2::Event::ParseError;
+use strict;
+use warnings;
+
+our $VERSION = '0.000012';
+
+BEGIN { require Test2::Event; our @ISA = qw(Test2::Event) }
```
exodist: Test2::Harness does not have the strict performance requirement of the Test2 dist, we can just use base/parent in events that should not move to Test2 itself.

lib/Test2/Event/TAP/Version.pm
```diff
@@ -0,0 +1,17 @@
+package Test2::Event::TAP::Version;
```
exodist: I feel like this event, or specifically one similar belongs in Test2... but I am not really sure.

lib/Test2/Formatter/EventStream.pm
```diff
@@ -35,16 +36,33 @@ sub init {
...
+
+    sub _event_to_json {
+        my ($self, $e) = @_;
+
+        my $json = eval { $J->encode($e) };
+        my $error = $@;
+        return $json if $json;
+
+        require Data::Dumper;
+        die "JSON encoding error: $error\n" . Data::Dumper::Dumper($e);
```
exodist: Why use Dumper ont he error instead of printing it directly? is it an object instead of a string?

lib/Test2/Harness/JSON.pm
```diff
@@ -0,0 +1,24 @@
+package Test2::Harness::JSON;
```
exodist: Nice splitting this out.

lib/Test2/Event/ParseError.pm
kentfredric: Given the drama surrounding base.pm and parent.pm, I'd say if you don't *need* to use either of them, and can avoid their use easily and elegantly, do so.

Using them simply makes the total complexity go up slightly without changing the effective behaviour.

lib/Test2/Harness/Parser/TAP/SubtestState.pm
```diff
@@ -0,0 +1,148 @@
...
+
+=item Chad Granum E<lt>exodist@cpan.orgE<gt>
+
+=back
+
+=head1 AUTHORS
+
+=over 4
+
+=item Chad Granum E<lt>exodist@cpan.orgE<gt>
```
exodist: You should add yourself here :-)

exodist:
If we are going through all this I would also really like to see Test2::Harness simply send the events to a formatter instead of implementing a renderer. The current Test2::Harness renderer would be ported to a formatter.

lib/Test2/Event/Encoding.pm
autarch: I will make a PR for the test-more repo to add this event and the `TAP::Version` event.

lib/Test2/Event/ParseError.pm
autarch: I have no strong opinion here. I was just copying what was in the `test-more` repo's event code.

autarch:
> If we are going through all this I would also really like to see Test2::Harness simply send the events to a formatter instead of implementing a renderer. The current Test2::Harness renderer would be ported to a formatter.

I don't think that makes sense. The difference in my mind is that a _renderer_ needs to be able to cope with multiple streams of events from multiple jobs being executed in parallel. IMO, we should not require this capability of formatters, as it adds a _lot_ of complexity and lot of state to manage.

lib/Test2/Harness/Parser/TAP/SubtestState.pm
autarch: BTW, we could use dzil to generate the`AUTHORS` and `CONTRIBUTORS` sections in all the pod.

autarch:
It also occurs to me that in the future we might want to split `Test2::Harness::Parser::TAP` into its own distro. This is the best TAP parser I've seen, and the only one that really tries to handle subtests at all. It'd be nice to make it available for others to use. There's really nothing about the parser that's specific to the Test2::Harness code except the base class from which it inherits, but that base class isn't doing much.

exodist:
You are right on both the renderer and the tap parser.

autarch:
I pushed a few more changes to this branch to address your comments. It will not pass CI until a new test-more is released.

Leont:
> It also occurs to me that in the future we might want to split Test2::Harness::Parser::TAP into its own distro. This is the best TAP parser I've seen, and the only one that really tries to handle subtests at all. 

Actually, my perl6 harness also tries to handle subtests (though that still a bit buggy currently), and has been doing so for two years. ;-)

> The difference in my mind is that a renderer needs to be able to cope with multiple streams of events from multiple jobs being executed in parallel.

Agreed

exodist:
I have merged this, but I am not ready to release it.

I used it to run Moose's test suite:

`yath -Ilib -j9 -LMoose`

In the middle of it:

```TAP
[LAUNCH]  _ : | : | : : | |  Started process with t/recipes/meta_globref_instancemetaclass.t
[ DIAG ]  + : | : | : : | |  undef
[PASSED]  | : = : | : : | |  t/basics/moose_respects_type_constraints.t passed
[LAUNCH]  | : _ : | : : | |  Started process with t/recipes/legacy_labeled_attributemetaclass.t
[ DIAG ]  | : + : | : : | |  undef
[PASSED]  | : | : | : : = |  t/basics/rebless.t passed
[LAUNCH]  | : | : | : : _ |  Started process with t/recipes/basics_bankaccount_methodmodifiersandsubclassing.t
[ DIAG ]  | : | : | : : + |  undef
[PASSED]  | : | : | : : | =  t/basics/always_strict_warnings.t passed
[LAUNCH]  | : | : | : : | _  Started process with t/recipes/basics_document_augmentandinner.t
[ DIAG ]  | : | : | : : | +  undef
[PASSED]  | : | : | = | | |  t/basics/super_warns_on_args.t passed
[LAUNCH]  | : | : | _ | | |  Started process with t/recipes/extending_debugging_baseclassrole.t
[ DIAG ]  | : | : | + | | |  undef
[PASSED]  | = | | | | | | |  t/basics/require_superclasses.t passed
[LAUNCH]  | _ | | | | | | |  Started process with t/recipes/basics_datetime_extendingnonmooseparent.t
[ DIAG ]  | + | | | | | | |  undef
[PASSED]  | | | = | | | | |  t/basics/super_and_override.t passed
[LAUNCH]  | | | _ | | | | |  Started process with t/recipes/meta_privateorpublic_methodmetaclass.t
[ DIAG ]  | | | + | | | | |  undef
[PASSED]  | | | | | | = | |  t/basics/create.t passed
[LAUNCH]  | | | | | | _ | |  Started process with t/recipes/roles_restartable_advancedcomposition.t
[ DIAG ]  | | | | | | + | |  undef
[PASSED]  | = | | | | | | |  t/recipes/basics_datetime_extendingnonmooseparent.t passed
[PASSED]  | : | | | | | = |  t/recipes/basics_bankaccount_methodmodifiersandsubclassing.t passed
[LAUNCH]  | : | | | | | _ |  Started process with t/recipes/roles_comparable_codereuse.t
[ DIAG ]  | : | | | | | + |  undef
[PASSED]  | | | | = | | | |  t/basics/global_destruction.t passed
[LAUNCH]  | | | | _ | | | |  Started process with t/recipes/basics_point_attributesandsubclassing.t
[ DIAG ]  | | | | + | | | |  undef
[PASSED]  | = | | | | | | |  t/recipes/basics_genome_overloadingsubtypesandcoercion.t passed
[LAUNCH]  | _ | | | | | | |  Started process with t/recipes/roles_applicationtoinstance.t
[ DIAG ]  | + | | | | | | |  undef
[PASSED]  | | = | | | | | |  t/recipes/legacy_labeled_attributemetaclass.t passed
[LAUNCH]  | | _ | | | | | |  Started process with t/recipes/meta_labeled_attributetrait.t
[ DIAG ]  | | + | | | | | |  undef
[PASSED]  | | | | | | | | =  t/recipes/basics_document_augmentandinner.t passed
[LAUNCH]  | | | | | | | | _  Started process with t/recipes/basics_binarytree_attributefeatures.t
[ DIAG ]  | | | | | | | | +  undef
[PASSED]  | | | | | = | | |  t/recipes/extending_debugging_baseclassrole.t passed
[LAUNCH]  | | | | | _ | | |  Started process with t/recipes/meta_table_metaclasstrait.t
[ DIAG ]  | | | | | + | | |  undef
[PASSED]  = | | | | | | | |  t/recipes/meta_globref_instancemetaclass.t passed
[LAUNCH]  _ | | | | | | | |  Started process with t/recipes/basics_company_subtypes.t
[ DIAG ]  + | | | | | | | |  undef
[PASSED]  | | | = | | | | |  t/recipes/meta_privateorpublic_methodmetaclass.t passed
[LAUNCH]  | | | _ | | | | |  Started process with t/recipes/extending_mooseish_moosesugar.t
[ DIAG ]  | | | + | | | | |  undef
[PASSED]  | | | | | | | = |  t/recipes/roles_comparable_codereuse.t passed
[LAUNCH]  | | | | | | | _ |  Started process with t/recipes/legacy_debugging_baseclassreplacement.t
[ DIAG ]  | | | | | | | + |  undef
[PASSED]  = | | | | | | | |  t/recipes/basics_company_subtypes.t passed
[LAUNCH]  _ | | | | | | | |  Started process with t/recipes/basics_http_subtypesandcoercion.t
[ DIAG ]  + | | | | | | | |  undef
```

I do not like seeing that 'undef', and I am pretty sure the previous version does not show it.

exodist:
yeah, with the old version that 'undef' is `Test file 't/recipes/meta_labeled_attributetrait.t' uses switches in the #! line, Falling back to IPC::Open3.`

So we definitely lost something here.

autarch:
I will investigate. I think it'll be easy enough to fix.

autarch:
I just pushed a fix to master.

PR #36, opened 2017-10-07
ProcMan: if long tests are waiting in line, bump them up

rjbs:


exodist:
Fixed in a different way.

Issue #73, opened 2018-05-09
Parsing error - slack messages are not being sent to correct channel


joevak:
Hi Chad,

The meta data for slack channels does not parse hyphens in the name.

`# HARNESS-META slack #employer-dev` leads to the error - `Failed to send slack message to #employer at Yath/Command/test.pm line 576.`

PR #75, opened 2018-05-28
fix for issue #73.

nawglan:
all of the harness directives expect a single word followed by
other stuff so do that first so we can treat 'meta' special.

meta expects 2 items, a key and a value.  they can be either white
space delimited, or dash delimited.

also, allow all directives to end with optional white space and a
comment.

add test to validate trailing comment, hyphenated keys and hyphenated
values for the meta harness directive.


nawglan:
need to remove the comment stripping part to allow slack channels.  will resubmit pull request after fixing this.

PR #88, opened 2018-08-24
Miscellaneous fixes around Formatter::Stream and the IPC runner

aeruder:
Several bugs around using the Stream formatter, specifically when the subprocess re-exec's itself.


exodist:
The test you wrote should go under a better name and be in the t2/ directory (tests run by yath when yaths own test suite runs). The test you replaced is intended to be a unit test only, this is more of an acceptance test. Rename+move the test. I am still looking for unintended side effects, but this is probably fine apart from the test.

aeruder:
Done

PR #113, opened 2019-08-27
Allow plugins to specify custom job runners on self-added files

ehuelsmann:
By specifying the 'via' attribute on instances of
Test2::Harness::Util::TestFile returned from $plugin->find_files(),
this change allows plugins to override the value selected by
default in Test2::Harness::Run::Runner.
I.e. this change allows specification of a value for '$task->{via}'
in Test2::Harness::Run::Runner::run_job.

### DIFF
--- a/lib/Test2/Harness/Util/TestFile.pm
+++ b/lib/Test2/Harness/Util/TestFile.pm
@@ -19,6 +19,7 @@ use Test2::Harness::Util::UUID qw/gen_uuid/;
 use Test2::Harness::Util::HashBase qw{
     -file -_scanned -_headers -_shbang -queue_args
     _category _stage
+    -via
 };
 
 *set_category = \&set__category;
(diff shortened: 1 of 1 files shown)


ehuelsmann:
Please provide your comment to the following thought: If we make the `via` attribute of `Test2::Harness::Util::TestFile` read/writable, then plugins can decide to set the attribute in the `munge_files` call. That would be good use of the existing plugin structure, but a rather obfuscated solution to a good use-case and would definitely call for a `How do I port my SourceHandler to Yath/Test2::Harness` type of manual.

ehuelsmann:
Ok. Then this PR by itself enables the 'via' accessor which provide the plugins with the current value for the TestFile instances as well as the code to pass the 'via' value on to the queue_item.

What remains is the work to create the plugin which uses this mechanism and the writing of documentation. The former is part of another project. Do you want the latter in this PR, or is it acceptable as it is?

ehuelsmann:
Ok. Thanks for the review. I'll withdraw the PR then; docs will need their own PR.

ehuelsmann:
With the above clarifications, I was able to build a working prototype
plugin+custom runner to run the Perl Cucumber tests (Test::BDD::Cucumber)
that I want to switch over to Yath.

I'll keep this issue updated with my status.

PR #115, opened 2019-09-06
Use new process groups for tests

exodist:


skaji:
You may want to check `$Config::Config{d_setpgrp}`.

exodist:
> You may want to check `$Config::Config{d_setpgrp}`.

I was wondering if that would be necessary. The perldoc seemed to indicate setpgrp(0,0) is portable, while other forms are not.

exodist:
I wonder if I should simplify it by making the parent call setpgrp on the child? Then I only have to use it in one place...

exodist:
@skaji force-pushed with a config check. No-op if the config is not set.

skaji:
If `setpgrp(0, 0)` is not called, then we should use `TERM` not `-TERM`.

exodist:
oh, good catch

exodist:
@skaji fixed.

skaji:
I'm not quite sure but
even if CAN_PGRP is true, `setpgrp(0, 0)` may not be called if `setpgrp => 1` ?

exodist:
@skaji there are some places that use run_cmd without setgrp, but all the PID's that we potentially kill with a '-' have the setgrp, with exception of Runner::Open3, which is only used on windows where setpgrp is not supported. So in theory the current state of the code lines up the -SIG's with setpgrp.

skaji:
@exodist If you are fine with it, then I'm fine with it 😉

exodist:
@skaji it does leave me conserned that it is an easy place to introduce bugs. Do you know of a simple way to check if a process is in the same process group as $$? Then I can add a conditional that would only add the '-' if the process I am about to kill is in a different group.

skaji:
If you are conserned, then I think yath should properly keeps whether setgrp is called or not in _PIDS and _JOBS together with pids.

aeruder:
My main concerns:

1. POSIX docs seems to suggest that `setpgrp(0, 0)` will always set the pgrp of the calling process to the same thing as `$$`.
2.`kill "-$sig", $pid` kills the pgrp `$pid`, not necessarily the pgrp of the process with pid `$pid`.
3. If you accept #1 then I think it would be fine to just always `kill "-$sig", $child_pid` and not have to worry about it being in the same pgrp as ourselves.
4. But if the `setpgrp(0, 0)` was skipped in the child process, it would just not send a signal to anything.

I think instead, the best thing to do is something along the lines of:
```
if (!kill("-$sig", "$child_pid_which_is_hopefully_also_child_pgrp")) { # kill will return 1 if signal successfully sent to pgrp - i.e. pgrp actually exists
  kill("$sig", "$child_pid");
}
```

exodist:
@aeruder #2 is a life saver if true. I was thinking there was a race condition (what if we tried to kill the process before the process could run setpgrp, if #2 is not true then we have a race condition. However if you are correct there is no race, and your code to kill the group and fall back to the process is perfect.

I am going to write a couple test scripts to verify.

exodist:
@aeruder confirmed:

```perl
use strict;
use warnings;

$SIG{INT} = sub { print "Parent got sigint\n"; exit 1; };
$SIG{TERM} = sub { print "Parent got sigterm\n"; exit 1; };

my $pid = fork();
unless($pid) {
    $SIG{INT} = sub { print "non-pgrp got sigint\n"; exit 1; };
    $SIG{TERM} = sub { print "non-pgrp got sigterm\n"; exit 1; };
    while (1) { sleep 1 }
    exit 0;
}

my $pgrp_pid = fork();
unless($pgrp_pid) {
    $SIG{INT} = sub { print "pgrp got sigint\n"; exit 1; };
    $SIG{TERM} = sub { print "pgrp got sigterm\n"; exit 1; };
    setpgrp(0,0);
    while (1) { sleep 1 }
    exit 0;
}

sleep 2;
print "\nSending -INT to both\n";
kill("-INT", $pid, $pgrp_pid);
sleep 2;
print "\nSending TERM to both\n";
kill("TERM", $pid, $pgrp_pid);
```

Results:

```
Sending -INT to both
pgrp got sigint

Sending TERM to both
non-pgrp got sigterm
```

exodist:
new force-push based on those findings @aeruder and @skaji   if this looks good now I will merge it

PR #116, opened 2019-09-08
Initial documentation for plugin entry points

ehuelsmann:
@exodist as promised, I've tried figuring out what each entry point in the plugin does. Let me know if you think this is a good start and which topics you want to be elaborated on or be struck (or moved to other modules).

Thanks for your time!

### DIFF
--- a/lib/App/Yath/Plugin.pm
+++ b/lib/App/Yath/Plugin.pm
@@ -23,3 +23,203 @@ sub claim_file {}
 sub inject_run_data {}
 
 1;
+
+__END__
+
+=pod
+
+=encoding UTF-8
+
+=head1 NAME
+
+App::Yath::Plugin - Parent of App::Yath plugins
+
+=head1 SYNOPSIS
+
+  package App::Yath::Plugin::SOME_PLUGIN;
+  use strict;
+  use warnings;
+
+  use parent 'App::Yath::Plugin';
+
+  sub find_files {
+    my @files;
+    # ...
+
+    return \@files;
+  }
+
+  1;
+
+  # which can then be used when invoking `yath`:
+
+  $ yath -pSOME_PLUGIN
+
+
+=head1 DESCRIPTION
+
+L<App::Yath>'s processing can be hooked into at specific points
+by plugins. These plugins can extend or change its default processing.
+
+This module provides an empty skeleton plugin. It's meant to be used
+as a parent class for any custom plugins.
+
+=head2 Instantiation
+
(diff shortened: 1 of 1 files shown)


lib/App/Yath/Plugin.pm
```diff
@@ -23,3 +23,203 @@ sub claim_file {}
...
+=head2 Instantiation
+
+In case a plugin has a C<new> function, it is assumed to be a constructor
+and an instance will be created using any parameters provided on the command
+line:
+
+  $ yath -pSOME_PLUGIN=abc,1,dev,delta
+
+  # causes the 'new' function to be invoked as:
+  #  App::Yath::Plugin::SOME_PLUGIN->new( abc => 1, dev => 'delta' );
```
exodist: You might make it clear that the arguments are passed as a list, so it is ok for a plugins new() method to take a list of args instead of named/hash-style args.

lib/App/Yath/Plugin.pm
```diff
@@ -23,3 +23,203 @@ sub claim_file {}
...
+
+
+=head1 FUNCTIONS
+
+=head2 options($command, $settings)
+
+This hook is called immediately after plugin instantiation.
+
+Returns a list of hash references to be added to the list of built-in
+options. Each hashref contains the following keys:
```
exodist: Would be nice to document what each of these does, or at the very least list which ones are required or not, that said this is better than no docs, so even adding a 'TODO' here works

lib/App/Yath/Plugin.pm
```diff
@@ -23,3 +23,203 @@ sub claim_file {}
...
+By returning a path (most often the C<$file> input parameter), this
+function stops default handling of C<$file>. This means that when
+C<$file> is a directory, subdirectories won't be searched. When C<$file>
+is a file path, that file won't be added to the list of test files
+handled by Yath. Presumably, the plugin has a mechanism to handle
+the C<$file> itself.
+
+=head2 block_default_search($settings)
+
+NOTE: Always a class method, even for instantiated plugins.
```
exodist: Hmm, I suspect this is a bug, but keep it in the docs until I fix it.

exodist:
Looks pretty good. I have a couple minor comments, nothing serious. I can merge as-is and tweak it, or you can tweak based on the comments before I merge.

ehuelsmann:
I'll enhance the documentation iteratively based on your review before merging, at least until the most prominent TODOs have been resolved to your satisfaction.

exodist:
Plugins are properly documented now in master.

PR #121, opened 2020-01-16
Skip retry for specific tests.

dkechag:
Adds the option --no-retry-long to avoid retry on all LONG tests and the #HARNESS-NO-RETRY directive which disables retry on a per-test basis.


exodist:
Please rewrite this patch against the major_refactor branch, which will be going stable in a couple weeks, if not sooner. I am not longer accepting patches against master.

dkechag:
OK, I'll redo it against the refactor at a later time, I see there are some changes there anyway that we will have to adapt our suite to.

PR #124, opened 2020-01-23
Provide some extra colors for jobs

atoomic:
Provide some extra colors, can see them by running
`perl -Ilib lib/Test2/Formatter/Test2.pm`

here is a sample using default terminal colors

Screen Shot 2020-01-22 at 5 11 14 PM


exodist:
Merged manually to remove the test logic, and move a comment.

atoomic:
thanks for considering the addition :-)

PR #126, opened 2020-01-30
[major_refactor] Consistent signal setup

atoomic:
This is a regression noticed with the 'major_refactor'
and signal improvements.

As shown by the provided test 'signals/abrt_or_iot.t',
this is flapping when using yath, whereas it's consistent
when using prove.

The root of the issue is the 'local %SIG = %SIG' which
has multiple issues.

1/ 'local %SIG' does not work, signals are not restored
when exiting the scope.

2/ '%SIG = %SIG' is also bad as it forces a hash setup
in a random order due to hash keys ordering...

Several signals are using multiple names for the same id:
- ABRT and IOT
- CHLD and CLD
- IO and POLL
- SYS and UNUSED

When relying on the signal name has shown by 'abrt_or_iot.t'
the first signal set is the one giving its name to the signal,
even if the signal is undefined...

By forcing the signal names to be ordered in the same way they
are listed in Config, we provide a consistent behavior and will
always use the same signal name.

### DIFF
--- a/lib/App/Yath/Command/runner.pm
+++ b/lib/App/Yath/Command/runner.pm
@@ -9,6 +9,7 @@ use goto::file();
 use Test2::Harness::IPC();
 
 use Carp qw/confess/;
+use Config qw/%Config/;
 use Scalar::Util qw/openhandle/;
 use List::Util qw/first/;
 use File::Path qw/remove_tree/;
@@ -56,7 +57,9 @@ sub generate_run_sub {
     my $runner_pid = $$;
     my $jump = setjump "Test-Runner" => sub {
         local $.;
-        local %SIG = %SIG;
+        my @SIGNAMES = grep { $_ ne 'ZERO' } split /\s+/, $Config{sig_name};
+        local @SIG{ @SIGNAMES } = @SIG{ @SIGNAMES };
+
         my $runner = $settings->build(
             runner => 'Test2::Harness::Runner',
 
--- a/t/integration/signals.t
+++ b/t/integration/signals.t
@@ -1,0 +1,24 @@
+use Test2::V0;
+
+use File::Temp qw/tempdir/;
+use File::Spec;
+
+use Test2::Harness::Util::File::JSONL;
+use App::Yath::Tester qw/yath/;
+
+my $dir = __FILE__;
+$dir =~ s{\.t$}{}g;
+
+for ( 1..10 ) {
+    # the tests are flapping when using something like '%INC = %INC'....
+    #   make sure the issue is fixed by running them a few times
+    my $out = yath(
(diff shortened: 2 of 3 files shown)


lib/App/Yath/Command/runner.pm
```diff
@@ -56,7 +57,9 @@ sub generate_run_sub {
     my $runner_pid = $$;
     my $jump = setjump "Test-Runner" => sub {
         local $.;
-        local %SIG = %SIG;
+        my @SIGNAMES = grep { $_ ne 'ZERO' } split /\s+/, $Config{sig_name};
+        local @SIG{ @SIGNAMES } = @SIG{ @SIGNAMES };
```
exodist: will local `$SIG{FOO} = $SIG{FOO}` will cause the 'FOO' key to exist where it previously did not, default value of undef. Does the value 'undef' have special meaning in %SIG? I am worried the 'undef' might override a default-to-ignore or similar.

I am wondering if we should skip using `local` and instead just copy %SIG, then copy it back and delete any new keys that were added since the copy was made.

atoomic: to be honest in the context of yath, we should probably consider to simply remove this line... 
why do we want to localize the signal here?

by default all signals exist and have some magic attached to them
are you worried about using a signal KEY (like FOO) not already defined?

What you are suggesting could work, but same as shown here you would have to make sure to define them in the correct order... 

The problem is for duplicate signals the first defined give its name to the signal.. 
So when setting/restoring them you need to be careful


Note: we are working on Perl patch to try to solve the issue at the root

atoomic: Note: I think the integration test is good to keep as it shows the regression / different behavior introduced and we want to avoid it if possible...

exodist: This is because the runner itself sets up signal handlers, and we want to be SURE they are cleared before any tests are run in the processes forked from the runner, which is how tests are run by default.

atoomic: ok, I still think that in this case the current patch is the best solution.
We are working for a Perl patch, but this will only make it to 5.32
https://github.com/Perl/perl5/pull/17509

Leont: > Does the value 'undef' have special meaning in %SIG? I am worried the 'undef' might override a default-to-ignore or similar.

Undef has a distinct side effect: it disables the C-level signal handler. That's exactly why `local %SIG` doesn't dwim, on restoration the correct perl-level handler is restored but the c-level handler isn't (as restoration doesn't trigger set magic)

exodist:
I merged your tests, but engineered a slightly different fix.

PR #127, opened 2020-01-30
Adjust SIG_MAP init

atoomic:
minor fixup

### DIFF
--- a/lib/Test2/Harness/IPC.pm
+++ b/lib/Test2/Harness/IPC.pm
@@ -18,7 +18,7 @@ BEGIN {
     my %SIG_MAP;
     my @SIGNAMES = split /\s+/, $Config{sig_name};
     my @SIGNUMS  = split /\s+/, $Config{sig_num};
-    while (@SIGNAMES || @SIGNUMS) {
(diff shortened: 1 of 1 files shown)


lib/Test2/Harness/IPC.pm
```diff
@@ -18,7 +18,7 @@ BEGIN {
     my %SIG_MAP;
     my @SIGNAMES = split /\s+/, $Config{sig_name};
     my @SIGNUMS  = split /\s+/, $Config{sig_num};
-    while (@SIGNAMES || @SIGNUMS) {
+    while (@SIGNAMES) {
```
atoomic: I understand that you want to empty both arrays... but if SIGNAMES is empty and SIGNUM is not...
you are going to use undef as hash key... and this would be a total other mess...
I think we can rely on Config being sane

exodist:
I think it would be better to get a list of keys from %sig beforehand, and then copy the values of those keys only. Once we want to restore %sig we set any keys from the initial list to the copy, but delete any keys that have come into existence since the copy. No need for config at all, and no accidental changes.

PR #129, opened 2020-01-30
[major_refactor] Add GitHub actions for Test2-Harness

atoomic:
Basic GitHub Actions integration so we can smoke the branch and detect regression over time

You can view a clean run here:
https://github.com/atoomic/Test2-Harness/actions/runs/33173879

Note that `integration` tests are removed on linux containers only as this causes infinite loop.
But the integration tests work fine on macOS containers...

You can see the infinite loop here from the Perl 5.30 run for example
https://github.com/atoomic/Test2-Harness/runs/418198841?check_suite_focus=true

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -1,0 +1,56 @@
+name: linux
+
+on:
+  push:
+    branches:
+      - '*'
+    tags-ignore:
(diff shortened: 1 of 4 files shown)


atoomic:
note this is also fixing a unit test on macOS

PR #131, opened 2020-02-04
[major_refactor] Make sure excluded files are excluded

atoomic:
Also add one integration test to confirm excluded files
are excluded as expected.

It's common to do something like this
    yath test --exclude-file t/boom.t t/*.t

### DIFF
--- a/lib/Test2/Harness/Finder.pm
+++ b/lib/Test2/Harness/Finder.pm
@@ -164,7 +164,7 @@ sub find_project_files {
         my $test;
         unless (first { $test = $_->claim_file($path, $settings) } @$plugins) {
             $test = Test2::Harness::TestFile->new(file => $path);
-            next unless @$input || $self->include_file($test);
+            next unless @$input && $self->include_file($test);
         }
 
         push @tests => $test;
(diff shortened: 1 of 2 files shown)


lib/Test2/Harness/Finder.pm
```diff
@@ -164,7 +164,7 @@ sub find_project_files {
         my $test;
         unless (first { $test = $_->claim_file($path, $settings) } @$plugins) {
             $test = Test2::Harness::TestFile->new(file => $path);
-            next unless @$input || $self->include_file($test);
+            next unless @$input && $self->include_file($test);
```
atoomic: this can also read as `if ! @$input || ! $self->include_file($test);`
in other way skip if include is not true

exodist:
This logic change is not correct. The @$input check was intended to make sure any file listed on the command line bypasses the 'include_file' check.

hmm, reading 'include_file' I think the real fix is to have include_file return an empty list when the file SHOULD be included, and return a reason a file is not included when it should nto be included. Then if @$input is true it should print a warning saying that the file was ont he command line, but was excluded because <reason>.  if @$input is empty then it jst does not include the file and moves on no warnings.

exodist:
Fixed in 6277bd6b

PR #135, opened 2020-02-05
[major_refactor] Add version to scripts/yath to run dzil build

atoomic:


exodist:
1) Why did you put it inside the $BEGIN instead of before/after it?
2) What problem does this solve? dzil build works just fine for me without this.

atoomic:
1. no special reason, the whole code was in a BEGIN block
but you are right should probably be something like `our $VERSION; BEGIN { $VERSION = ... }`
2. `dzil build` was failing 
```
╰─> dzil build
[DZ] beginning to build Test2-Harness
[DZ] guessing dist's main_module is lib/Test2/Harness.pm
[RewriteVersion] Skipping: no "our $VERSION = '...'" found in "scripts/yath"
[DZ] writing Test2-Harness in Test2-Harness-1.000000
Error open (<:unix) on '/Users/nicolas/workspace/perl-modules/Test2-Harness/Test2-Harness-1.000000/cpanfile': No such file or directory at /usr/local/perl/perls/perl-5.26.2/lib/site_perl/5.26.2/Dist/Zilla/File/OnDisk.pm line 34.
```

exodist:
That's just a note in the output, I get it too, but it does not block the build. I would focus in your file not found error, that one is new to me.

atoomic:
let's close it, a `git clean -df` solved the issue lol

PR #136, opened 2020-02-06
[major_refactor] Do not add '.' to @INC by default.

atoomic:
Use sane default values for @INC.

The default behavior should be the sane and recommended
one, all the most it matches Perl default values.

This is great that we have an option to add '.' in @INC
when needed but this should not be the default.

### DIFF
--- a/lib/App/Yath/Options/Runner.pm
+++ b/lib/App/Yath/Options/Runner.pm
@@ -62,7 +62,7 @@ option_group {prefix => 'runner', category => "Runner Options"} => sub {
     option unsafe_inc => (
         description => "perl is removing '.' from \@INC as a security concern. This option keeps things from breaking for now.",
         env_vars    => [qw/PERL_USE_UNSAFE_INC/],
-        default     => 1,
+        default     => 0,
     );
 
     option preloads => (


exodist:
`prove` defaults to include '.' in @inc. There are still too many test suits on cpan that depend on this behavior for it to default to 'off'. Removing it by default right now could prevent adoption of yath.

atoomic:
I understand your point of view, but on the other side having better
default value is another argument for adopting `yath` instead of `prove`.

All the most if you want the legacy behavior, you have one argument for it.

We have patched several modules among the hundreds we used from CPAN, 
and do not think that many are still relying on 'dot in inc'.

toddr:
> I understand your point of view, but on the other side having better
> default value is another argument for adopting `yath` instead of `prove`.

This is an interesting argument. If someone is adopting yath anew. Why not encourage good behavior on initial use?

exodist:
Its a tough balance. The people who really really want to make things better will fix their code. But most programmers are lazy and will snidely report the lack of '.' Breaking their tests as a bug and proudly proclaim they will not use yath until the "bug" is "fixed", then will go on to tell the rest of the world that the perl community is an elitist group that claims they support backwords compatibility while in reality breaking things in a whim.

I wish this were not the case, but experience tells me it is unavoidable. The  again being stuck with 'prove' is a fitting consequence for such people, or would be if they did not derive pleasure from having excuses to stay in the stone age.

exodist:
You know what, screw the haters, lets do things right the first time.

toddr:
Keeping in mind we're just changing the default. they can always set the env var or pass it on command line.

PR #137, opened 2020-02-06
[major_refactor] Do not use relative path in @INC

atoomic:
Using relative path in @INC could lead to unexpected behavior.
Code can be loaded from uncontrolled locations.

When setting @INC with devlib: 'lib', 'blib', 'blib/arch'
we should not use the relative PATH but prefer the absolute one.

Note that the workardound for 'projects' command need some extra love,
as this seems to be the expected behavior for projects, we should avoid
using an environment variable.

note: built on top of #136

### DIFF
--- a/lib/App/Yath/Command/runner.pm
+++ b/lib/App/Yath/Command/runner.pm
@@ -20,7 +20,7 @@ use Test2::Util qw/clone_io/;
 
 use Long::Jump qw/setjump longjump/;
 
-use Test2::Harness::Util qw/mod2file write_file_atomic open_file/;
+use Test2::Harness::Util qw/mod2file write_file_atomic open_file clean_path/;
 
 use Test2::Harness::Util::IPC qw/swap_io/;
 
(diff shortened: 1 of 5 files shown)


exodist:
This PR still has some '.' in @INC pr code in it, can you remove it so the 2 PR's are fully separate?

lib/App/Yath/Options/Runner.pm
```diff
@@ -62,7 +62,7 @@ option_group {prefix => 'runner', category => "Runner Options"} => sub {
     option unsafe_inc => (
         description => "perl is removing '.' from \@INC as a security concern. This option keeps things from breaking for now.",
         env_vars    => [qw/PERL_USE_UNSAFE_INC/],
-        default     => 1,
+        default     => 0,
```
exodist: This is part of your other PR, can you pull this out of this PR?

atoomic: sure will do, this will increase the test complexity but can work around it

atoomic: updated via ea721431

PR #138, opened 2020-02-06
[major_refactor] use perldoc to display command help

atoomic:
Experimental idea to use perldoc to display commmand help usages.

When possible use perldoc to display contextual help.

Note: need to test on a distro where pod are generated :-)


exodist:
I appreciate the idea, but would prefer not to go this route.

atoomic:
no problems thanks for your answer

PR #139, opened 2020-02-06
[major_refactor] Preserve test --retry meaning

atoomic:
We should not need to set '--retry=2' to retry a test.
The previous implementation would be more accurate if we
rename '--retry' '--run-up-to-x-times'.

As documented:
"retry=1 means failing tests will be attempted twice"

--retry and --retry=1 should be similar and should request for
a job to be run one more time on failure.

--retry=0 is similar to --no-retry
--retry=1 try an extra time if it fails, which can run the test
up to two times
--retry=2 try to rerun a failing test up to twice, so up to three runss
...


toddr:
Yep. in the current implementation, I'm not clear the difference betwen --retry=0 and --retry=1

PR #141, opened 2020-02-07
[major_refactor] Use recommended numeric comparison for $]

atoomic:
This was recently updated during Perl 5.32 development cycle.

Upstream-URL: https://github.com/Perl/perl5/commit/601390ce200db31ae8eecba5649a8c1f2e14ca3d

### DIFF
--- a/t/integration/test.t
+++ b/t/integration/test.t
@@ -173,7 +173,7 @@ yath(
     );
 }
 
-if ($] ge '5.026000') {
+if ("$]" >= 5.026) {
     note q[Checking %INC and @INC setup];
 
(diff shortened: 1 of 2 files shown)


exodist:
2 notes

1: HashBase is a generated file, to change that file you would need to update the Object::HashBase distribution and regenerate

2: the string comparison is what was recommended in older perls, does this change break things on older perls?

atoomic:
1. going to revert the change for this file
2. yes the string comparison work with all perl versions, this is what we are using in Devel::PPPort
going to resubmit

PR #143, opened 2020-02-08
[major_refactor] Improve durations.json file from speedtag

atoomic:
Use relative path in the json file and use
one line per file so we can easily grep the test
in the durations file.

Tests are also sorted.


exodist:
I think 'pretty' should be optional, maybe a CLI arg for the speedtag command.

atoomic:
ok going to add `--pretty` to the cli

atoomic:
pull request updated with `--pretty` option which default to false

PR #148, opened 2020-02-11
[major_refactor] Do not add missing absolute path to @INC

### LINKED ISSUE #147: yath should default things like --blib based on -d ./blib
I've just done a quick parse of `yath test --help`.

I suggest that --blib, --tlib, --lib, and --dev-lib default to not injecting into @INC if the directory they would inject is not a valid directory path (symlinks are ok?)

atoomic:
Fixes #147:

### DIFF
--- a/lib/App/Yath/Command/runner.pm
+++ b/lib/App/Yath/Command/runner.pm
@@ -277,7 +277,10 @@ sub build_init_state {
 
     # Make sure our specified includes are at the front of the list.
     my %seen;
-    @INC = grep { !$seen{$_}++ } $job->unsafe_inc ? ('.') : (), map { clean_path( $_ ) } $job->includes, @INC;
+    @INC = grep { !$seen{$_}++ } $job->unsafe_inc ? ('.') : (), map {
+        my $p = clean_path( $_ );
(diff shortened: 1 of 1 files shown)


lib/App/Yath/Command/runner.pm
```diff
@@ -277,7 +277,10 @@ sub build_init_state {
 
     # Make sure our specified includes are at the front of the list.
     my %seen;
-    @INC = grep { !$seen{$_}++ } $job->unsafe_inc ? ('.') : (), map { clean_path( $_ ) } $job->includes, @INC;
+    @INC = grep { !$seen{$_}++ } $job->unsafe_inc ? ('.') : (), map {
+        my $p = clean_path( $_ );
+        index($p, '/', 0 ) != 0 || -d $p ? $p : ()
```
atoomic: this can probably be converted to a simple `-d $p ? $p : ()` as clean_path is going to guarantee an absolute PATH

exodist:
See discussion in #147 and commit https://github.com/Test-More/Test2-Harness/commit/7833ab76063a99dd940e9454c1a37400cef92ceb

Issue #150, opened 2020-02-13
yath test --no-show-times


atoomic:
I was trying to reduce the verbosity of `yath` command using `--qvf` and friends

I came up with something like this 
`yath test --qvf --no-show-times --no-show-job-end --no-progress t`

I'm surprise that `--no-show-times` do not disable the `TIME` output

```
(  TIME  )  job  2    Startup: 0.01391s | Events: 0.00000s | Cleanup: 3.57830s | Total: 3.59221s
(  TIME  )  job  3    Startup: 0.00622s | Events: 0.00000s | Cleanup: 3.63381s | Total: 3.64003s
(  TIME  )  job  1    Startup: 0.02210s | Events: 0.00000s | Cleanup: 3.81469s | Total: 3.83679s
(  TIME  )  job  8    Startup: 0.08472s | Events: 0.18937s | Cleanup: 0.00941s | Total: 0.28350s
(  TIME  )  job  4    Startup: 0.00832s | Events: 0.00000s | Cleanup: 3.71259s | Total: 3.72091s
(  TIME  )  job  9    Startup: 0.08413s | Events: 0.05473s | Cleanup: 0.00729s | Total: 0.14615s
(  TIME  )  job  5    Startup: 0.00391s | Events: 0.00000s | Cleanup: 3.95473s | Total: 3.95864s
(  TIME  )  job  6    Startup: 0.01928s | Events: 0.00000s | Cleanup: 4.06210s | Total: 4.08138s
```

Either I've a bad understand of the --no-show-times option
or we could consider an additional flag to disable the time?
or there is a bug and we should also disable these TIME output lines using the existing arg

not sure, need to investigate

exodist:
Hmm, there is more to this than meets the eye.

`yath test t2`  does not show me the times
`yath test t2 --qvf`does show me the times.

So the problem here is actually in the QVF renderer.

exodist:
In the QVF renderer it sets verbosity to 100, if I comment that out it no longer shows times. Which means the QVF renderer is not being as smart as it needs to be.

PR #151, opened 2020-02-13
[major_refactor] Disable TIME output when using --no-show-times

atoomic:
Fixes GH #150


exodist:
Fixed in https://github.com/Test-More/Test2-Harness/commit/049d92d8a6da6dd13abb7bf0cc50f92c52184dea instead

atoomic:
cool thanks!

PR #164, opened 2020-03-08
Fix -j option

### LINKED ISSUE #163: -j issue
With the refactored yath, when I start with -j, the correct number of parallel jobs is loaded with LONG tests, but once one of them finishes I am down to 1 less parallel job. So -j4 runs 3 after one is finished etc.
Am I the only one seeing this? If so I will debug it on my system...

dkechag:
Closes #163 

Using the -j option, the first time a task finished, a `$state->advance` was done without a `$state->next_task`, so a parallel thread was essentially lost.

I added a quick test based on the existing one, although it is not nice (it makes an assumption that the tests used cant go from start to exit immediately), so maybe a nicer one should be written?

PS. The -j/jobs nomenclature (instead something like threads/parallel etc) is a bit unfortunate I realized when reading the code, but it was inherited from prove of course ;)

### DIFF
--- a/lib/Test2/Harness/Runner.pm
+++ b/lib/Test2/Harness/Runner.pm
@@ -405,6 +405,7 @@ sub next {
 
     my $state = $self->state;
 
+    OUTER:
     while (1) {
         if(my $task = $state->next_task()) {
             next unless $task->{stage} eq $self->{+STAGE};
@@ -413,7 +414,7 @@ sub next {
 
         if (my $lock = $self->lock_dispatch) {
             while (1) {
-                next if $state->advance();
+                next OUTER if $state->advance();
                 last;
             }
         }
--- a/t/integration/concurrency.t
+++ b/t/integration/concurrency.t
@@ -46,7 +46,7 @@ yath(
         @order = map { $_->[0] } sort { $a->[1] <=> $b->[1] } @order;
 
 # The first 4 events should be starts since we have 4 concurrent jobs
-# After they start we MUST see and exit before any more can start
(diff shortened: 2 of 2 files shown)


lib/Test2/Harness/Runner.pm
```diff
@@ -413,7 +414,7 @@ sub next {
 
         if (my $lock = $self->lock_dispatch) {
             while (1) {
```
exodist: This change essentially makes the nested while loop unnecessary. removing the `while` and the `last` would essentially be the same change. This makes me unsure if this fix might introduce a bug. I have to re-read the code and see if I can figure out why that second loop is there.

exodist:
@dkechag I took the test you wrote, but made an alternate solution in the same area where your solution was written. You were correct about the bug, but the solution you wrote missed the fact that the while loop was intended to let the process advance as many things as it could while it had the lock. The solution I added fixes the bug, but keeps the desired loop behavior.

exodist:
Please verify this fixes the problem on your end, assuming it does I can mint a new release.

dkechag:
Yes, your commit seems to work fine, thanks.
Completely unrelated, but to not messaging separately: you have left one mention of `HARNESS-CATEGORY-LONG` in the doc, which is what was proposed originally instead of `HARNESS-DURATION-LONG`. Do you intend for both to continue work interchangeably? If yes, maybe add a mention of the synonym under `HARNESS-DURATION-LONG` (that CATEGORY-LONG is the same), otherwise replace the mention of CATEGORY-LONG?

PR #177, opened 2020-04-07
Fixed - yath shebang set on install

eilara:
When set as `#!/usr/bin/env perl` the install process does not touch the shebang line in the yath script. But people expect the installer to replace it with the Perl used by the installer.

This tiny PR sets shebang on yath to `#!/usr/bin/perl`, which gets replaced by the installer with the correct `perl`.


exodist:
Huh, I thought that was fixed at some point to make `/usr/bin/env perl` also get modified. I will merge this later this evening and do a new release, cannot do it quite yet.

PR #182, opened 2020-04-21
replay: allow a relative filename instead of job id

rjbs:


toddr:
What is the goal here? Is this for an addon? I'm vaguely concerned it will impact how Test2::Harness::Renderer::JUnit processes re-runs.

exodist:
@toddr this is for the 'replay' command, which lets you use a log file as input and replay the entire test run from the log instead of running the tests again. The command lets you list job-id's to only show the specific job-ids and not show any output for the other jobs. @rjbs is working on a PR that lets you specify filenames instead of only job-ids as that is easier for humans. This should have no effect on re-runs.

PR #183, opened 2020-04-21
yath failed: do not die on zero failures

rjbs:


exodist:
I would be very curious to see what you ran to make this necessary.

rjbs:
I ran "yath failures" on a jsonl log with no failures in it.  That's all!

rjbs:
To elaborate just a tiny bit:  I am writing a very small amount of automation that injects a program into a test VM, uses that program to run the test suite, and then I extract the jsonl file for analysis by the automation program, which then uploads a report to a shared workspace.

Part of that analysis is "see whether there are failures," which it does by running yath failures.  But that dies if there are none!

Even if the simplest thing is "use some other mechanism to determine that," it seems like dying on a zero-failure log is just sub-optimal behavior.

exodist:
Thats what I get for looking at this on my phone, I thought this was for the test command, haha.

PR #189, opened 2020-07-09
Fix permissions on temp dirs

### LINKED ISSUE #185: TMPDIR is taken over, but weird permissions are used
Test::Harness::Runner::Job sets up environment variables for the jobs it will run.  Among these, it sets TMPDIR to `$self->tmp_dir`.  That's computed from this:

```perl
File::Temp::tempdir("XXXXXX", DIR => $self->runner->tmp_dir);
```

File::Temp::tempdir boils down to calling its internal _gettemp, which will do this:

```perl
} elsif ($options{"mkdir"}) {

  # Open the temp directory
  if (mkdir( $path, 0700)) {
    # in case of odd umask
    chmod(0700, $path);

    return undef, $path;
```

So, the TMPDIR we set is set `0700`, but a normal unix TMPDIR is usually `1777`.  Also, the *job's* tmp_dir is under the *runner's* tmp_dir, which is another `0700` directory.

The net result is that processes spawned by tests can't be guaranteed that they can read and write in the TMPDIR.

For a more concrete example:  I have a set of tests that runs as root and then forks a subprocess, which switches to a lower-privilege user.  The subprocess then wants to create a lockfile using `Path::Tiny->tempfile`, but:

1. TMPDIR is 0700, owned by root, so the subprocess may not write there
2. a parent of TMPDIR (the runner's tmp_dir) is *also* 0700, so chmoding TMPDIR before dropping privileges won't help

I'm not sure what I actually suggest here, but I'm going to start by saying that my current thinking is that *either* the job's tmp_dir is for the job's use and shouldn't take over TMPDIR *or* if it's meant to create a localized TMPDIR-like space, it should make sure that it's 1777 and that every parent of it have all of the bits in 0555 set.

exodist:
Fixes #185

### DIFF
--- a/lib/App/Yath/Command/test.pm
+++ b/lib/App/Yath/Command/test.pm
@@ -333,6 +333,7 @@ sub build_run {
     my $run = $settings->build(run => 'Test2::Harness::Run');
 
     mkdir($run->run_dir($dir)) or die "Could not make run dir: $!";
+    chmod(1777, $dir) or warn "Could not chmod run dir: $!\n";
 
     return $self->{+RUN} = $run;
 }
--- a/lib/App/Yath/Options/Workspace.pm
+++ b/lib/App/Yath/Options/Workspace.pm
@@ -46,6 +46,7 @@ option_group {prefix => 'workspace', category => "Workspace Options"} => sub {
             }
             else {
                 mkdir($workdir) or die "Could not create workdir: $!";
+                chmod(1777, $workdir) or warn "Could not chmod work dir: $!\n";
             }
 
             return;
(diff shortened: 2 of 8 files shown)


rjbs:
Short answer:  It didn't solve the problem.  I ended up with

```
Caught exception in subtest: Failed to start login handshake: $VAR1 = '500 Error in tempfile() using template /tmp/yath-229591-f2a_3Q/tmp/kKgGVq/XXXXXXXXXX: Could not create temp file /tmp/yath-229591-f2a_3Q/tmp/kKgGVq/6Unq1DqbCZ: Permission denied
```

I did not dig deeper than that, yet.  The key thing to know is that yath is being run as root (not ideal, but necessary for now) and the individual test changes effective user as part of testing a service that runs as that user.

exodist:
Hmm. I will probably still merge this as an improvement. But I will
probably add an option to disable run/job specific temp dirs that you can
use in cases like this. Probably also a directive you can put inside
specific test files that need to use the system temp.

PR #201, opened 2020-11-06
Explicitly close some file handles

atoomic:
After updating to v1.000038 from v1.000000 we can start noticing some `Too many open files` issues.
I did a quick review of the change from `v1.000000..v1.000038` and check all new introduced `~open` 
and add an extra explicit close for them

### DIFF
--- a/lib/App/Yath/Command/runner.pm
+++ b/lib/App/Yath/Command/runner.pm
@@ -374,6 +374,9 @@ sub update_io {
         my $msg = "$_[0] at $caller[1] line $caller[2] ($caller2[1] line $caller2[2]).\n";
         print $stderr $msg;
         print STDERR $msg;
+        close($out_fh);
+        close($err_fh);
+        close($in_fh) if $in_fh;
         POSIX::_exit(127);
     };
 
--- a/lib/Test2/Harness/Util.pm
+++ b/lib/Test2/Harness/Util.pm
@@ -318,6 +318,7 @@ sub find_libraries {
                 next unless -f $path && $path =~ m/\.pm$/;
                 push @found => [$path, $prefix];
             }
+            closedir($dh);
         }
 
         push @bases => $new_base if @$new_base;
--- a/lib/App/Yath/Command/spawn.pm
+++ b/lib/App/Yath/Command/spawn.pm
@@ -180,10 +180,12 @@ sub run {
             sleep 0.2;
         }
     }
+    close($wfh);
 
     $self->clear_sig_handlers();
 
     my $exit = read_line($fh) // die "Could not get exit code";
+    close($fh);
     $exit = parse_exit($exit);
     if ($exit->{sig}) {
         print STDERR "Terminated with signal: $exit->{sig}.\n";
--- a/lib/App/Yath/Tester.pm
+++ b/lib/App/Yath/Tester.pm
@@ -129,6 +129,7 @@ sub yath {
             push @lines => @new;
             print map { chomp($_); "DEBUG: > $_\n" } @new if $debug > 1;
         }
+        close($rh);
     }
     else {
         print "DEBUG: Waiting for $pid\n" if $debug;
--- a/lib/Test2/Harness/Collector.pm
+++ b/lib/Test2/Harness/Collector.pm
@@ -170,6 +170,7 @@ sub process_runner_output {
             stream => Test2::Harness::Util::File::Stream->new(name => File::Spec->catfile($auxdir, $path)),
         };
     }
+    closedir($dh);
 
     for my $file (sort keys %{$self->{+RUNNER_AUX_HANDLES}}) {
         my $data   = $self->{+RUNNER_AUX_HANDLES}->{$file};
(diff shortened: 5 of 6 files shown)


lib/App/Yath/Command/runner.pm
```diff
@@ -374,6 +374,11 @@ sub update_io {
         my $msg = "$_[0] at $caller[1] line $caller[2] ($caller2[1] line $caller2[2]).\n";
         print $stderr $msg;
         print STDERR $msg;
+        eval {
```
exodist, line 377: This eval is not necessary, close() is not fatal when it fails. IF we do keep the eval I would like $@ localized just before the eval.

atoomic: fair enough let's remove the eval :-)

atoomic: updated via 3946062e5efa2536b70530bc0a2c71df74cf4fda

exodist:
Can you explain to me why this is necessary, with exception of $wfh in one of the files, most of these look like the scope should end the filehandle anyway, so they seem unnecessary. Am I wrong about when filehandles are cleaed up, is it at sub exit instead of scope exit?

atoomic:
@exodist IMO the fh should also close when exiting the scope. So yes most of them are not necessary.
As updated in the description, after updating to v1.000038 from v1.000000 we can start noticing some `Too many open files issues`. I did a quick pass to check all added `open` during these iterations and ensure they are close as soon as possible.

I agree some close are not necessary, but I also realized that the existing code was doing a pretty good job of using explicit close when needed, so I decided to add them too.

exodist:
ah, that makes sense. Did this patch solve your "too many handles" issue?

atoomic:
I'm currently working on this right now, the answer is not obvious as this is a flapping issue... that we start noticing since the update... so might need a few extra days to confirm if it fixes it or not

exodist:
ok, either way I have no issues with this patch, I think it should be a no-op, explicit close is fine though too. But if you are still working on the problem I will keep it open for any additional patches you may want to attach.

atoomic:
this is not good enough, I can still see some random issues while running our test suite
need to keep investigating

toddr:
I wonder if there's a refcount issue introduced recently that might be the source of the problem. is something keeping track of dead handles in a way that's maybe not getting freed? 

FTR we have about 4000 test files and are running at -j12

toddr:
just got:

```
19:31:42  Could not open events dir: Too many open files at Test2/Harness/Collector/JobDir.pm line 436, <$fh> line 1.
```

toddr:
Actually the problem happened on 1.000000. So this may go back to a problem between 0.999009 and 1.000000

exodist:
Are open files limited by user, process, or system?

toddr:
On most Linux systems, the default for compiled binaries is a max of 1024 file handles per process.

toddr:
Don't worry about this for now. I have a way to find the root cause on monday. it should be easy to detect. I just need to check /proc/$$/fd and find out which parent yath process is hoarding file descriptors since they're being passed to the children on fork.

toddr:
I assume if I can tell you what files aren't closing, then it'll be easy to determine the root cause?

exodist:
Potentially. I would focus on the collector and the runner, they are the ones that need the most filehandles. The runner creates the handles for the children when it spawns them, but should let them go right after. The collector opens all the files produced by tests, but should let them go when a test completes.

My hunch is that either the collector is not letting go of file handles when it is done with them (unexpected references kept?), Or with 12 processes running there is a combination of 12 of your tests that combined with what yath uses result in too many files. Maybe even 1 or 2 filehandle heavy tests?

atoomic:
Thanks @exodist for these details.
This is an issue from the collector when we run too many tests / too fast...
The collector, just try to open every files without any limits...

We could limit the number of jobs opened to avoid(/fix) the issue
Note that when using retry the collector is also keeping open FH... 
this is probably something we can ignore for now...

atoomic:
The real fix is #203

lib/Test2/Harness/Util.pm
```diff
@@ -318,6 +318,7 @@ sub find_libraries {
                 next unless -f $path && $path =~ m/\.pm$/;
                 push @found => [$path, $prefix];
             }
+            closedir($dh);
```
atoomic, line 321: useless

lib/App/Yath/Command/runner.pm
```diff
@@ -374,6 +374,9 @@ sub update_io {
         my $msg = "$_[0] at $caller[1] line $caller[2] ($caller2[1] line $caller2[2]).\n";
         print $stderr $msg;
         print STDERR $msg;
+        close($out_fh);
+        close($err_fh);
+        close($in_fh) if $in_fh;
```
atoomic, line 379: useless

atoomic (commented):
pointed the useless `close|closedir` which should be removed from that PR
leaving a few valuable earlier closer
need to adjust and resubmit

PR #203, opened 2020-11-09
Avoid opening too many jobs in the collector. Only parse so many at o…

toddr:
…ne time.

Most unix systems limit the number of file handles any one process can
have open at a time to 1024. When the collector gets behind, it may try
to open too many files for completed jobs at once. Limit this with a
constant. They'll be picked up once the ones polled are processed.

### DIFF
--- a/lib/App/Yath/Command/collector.pm
+++ b/lib/App/Yath/Command/collector.pm
@@ -6,6 +6,7 @@ our $VERSION = '1.000039';
 
 use File::Spec;
 
+use App::Yath::Options;
 use App::Yath::Util qw/isolate_stdout/;
 
 use Test2::Harness::Util::JSON qw/decode_json/;
@@ -15,6 +16,12 @@ use Test2::Harness::Run;
 
 use parent 'App::Yath::Command';
 use Test2::Harness::Util::HashBase;
+
+include_options(
+    'App::Yath::Options::Debug',
+    'App::Yath::Options::PreCommand',
+    'App::Yath::Options::Collector',
+);
 
 sub internal_only   { 1 }
 sub summary         { "For internal use only" }
@@ -28,7 +35,7 @@ sub run {
 
     my $fh = isolate_stdout();
 
-    my $settings = Test2::Harness::Settings->new(File::Spec->catfile($dir, 'settings.json'));
+    my $settings = $self->settings;
 
     require(mod2file($collector_class));
 
--- a/lib/App/Yath/Command/collector.pm
+++ b/lib/App/Yath/Command/collector.pm
@@ -6,7 +6,6 @@ our $VERSION = '1.000039';
 
 use File::Spec;
 
-use App::Yath::Options;
 use App::Yath::Util qw/isolate_stdout/;
 
 use Test2::Harness::Util::JSON qw/decode_json/;
@@ -16,12 +15,6 @@ use Test2::Harness::Run;
 
 use parent 'App::Yath::Command';
 use Test2::Harness::Util::HashBase;
-
-include_options(
-    'App::Yath::Options::Debug',
-    'App::Yath::Options::PreCommand',
-    'App::Yath::Options::Collector',
-);
 
 sub internal_only   { 1 }
 sub summary         { "For internal use only" }
@@ -35,7 +28,7 @@ sub run {
 
     my $fh = isolate_stdout();
 
-    my $settings = $self->settings;
+    my $settings = Test2::Harness::Settings->new(File::Spec->catfile($dir, 'settings.json'));
 
     require(mod2file($collector_class));
 
--- a/lib/Test2/Harness/Collector.pm
+++ b/lib/Test2/Harness/Collector.pm
@@ -24,6 +24,8 @@ use Test2::Harness::Util::HashBase qw{
     <run_dir
     <runner_pid +runner_exited
 
+    <backed_up
+
     +runner_stdout +runner_stderr +runner_aux_dir +runner_aux_handles
 
     +task_file +task_queue +tasks_done +tasks
@@ -52,6 +54,8 @@ sub init {
 sub process {
     my $self = shift;
 
+    my $settings = $self->settings;
+
     while (1) {
         my $count = 0;
         $count += $self->process_runner_output if $self->{+SHOW_RUNNER_OUTPUT};
@@ -66,7 +70,7 @@ sub process {
 
         while(my ($job_try, $jdir) = each %$jobs) {
             my $e_count = 0;
-            for my $event ($jdir->poll(1000)) {
+            for my $event ($jdir->poll($self->settings->collector->max_poll_events // 1000)) {
                 $self->{+ACTION}->($event);
                 $count++;
             }
@@ -76,7 +80,7 @@ sub process {
             my $done = $jdir->done or next;
 
             delete $jobs->{$job_try};
-            unless ($self->settings->debug->keep_dirs) {
+            unless ($settings->debug->keep_dirs) {
                 my $job_path = $jdir->job_root;
                 # Needed because we set the perms so that a tmpdir under it can be used.
                 # This is the only remove_tree that needs it because it is the
@@ -97,7 +101,7 @@ sub process {
 
     $self->{+ACTION}->(undef) if $self->{+JOBS_DONE} && $self->{+TASKS_DONE};
 
-    remove_tree($self->{+RUN_DIR}, {safe => 1, keep_root => 0}) unless $self->settings->debug->keep_dirs;
+    remove_tree($self->{+RUN_DIR}, {safe => 1, keep_root => 0}) unless $settings->debug->keep_dirs;
 
     return;
 }
@@ -214,6 +218,29 @@ sub process_tasks {
     return $count;
 }
 
+sub send_backed_up {
+    my $self = shift;
+    return if $self->{+BACKED_UP}++;
+
+    # This is an unlikely code path. If we're here, it means the last loop couldn't process any results.
+    my $e = $self->_harness_event(0, undef, time, info => [{details => <<"    EOT", tag => "INTERNAL", debug => 1, important => 1}]);
+*** THIS IS NOT FATAL ***
+
+ * The collector has reached the maximum number of concurrent jobs to process.
+ * Testing will continue, but some tests may be running or even complete before they are rendered.
+ * All tests and events will eventually be displayed, and your final results will not be effected.
+
+Set a higher --max-open-jobs collector setting to prevent this problem in the
+future, but be advised that could result in too many open filehandles on some
+systems.
+
+This message will only be shown once.
+    EOT
+
+    $self->{+ACTION}->($e);
+    return;
+}
+
 sub jobs {
     my $self = shift;
 
@@ -221,13 +248,13 @@ sub jobs {
 
     return $jobs if $self->{+JOBS_DONE};
 
-    # Don't monitor more than 'max_jobs_to_process' or we might have too many open file handles and crash
-    # Max open files handles on a process applies. Usually this is 1024 so we can't have everything open at once when we're behind.
-    my $max_jobs_to_process = $self->settings->collector->max_jobs_to_process // 1024;
-    my $additional_jobs_to_parse = $max_jobs_to_process - scalar keys %$jobs;
+    # Don't monitor more than 'max_open_jobs' or we might have too many open file handles and crash
+    # Max open files handles on a process applies. Usually this is 1024 so we
+    # can't have everything open at once when we're behind.
+    my $max_open_jobs = $self->settings->collector->max_open_jobs // 1024;
+    my $additional_jobs_to_parse = $max_open_jobs - keys %$jobs;
     if($additional_jobs_to_parse <= 0) {
-        # This is an unlikely code path. If we're here, it means the last loop couldn't process any results.
-        print "WARN: max jobs still exceded. Continuing to process jobs\n";
+        $self->send_backed_up;
         return $jobs;
     }
 
@@ -282,9 +309,7 @@ sub jobs {
     }
 
     # The collector didn't read in all the jobs because it'd run out of file handles. We need to let the stream know we're behind.
-    if( scalar keys %$jobs >= $max_jobs_to_process ) {
-        print STDERR "WARN: The Yath Collector is running behind. More than $max_jobs_to_process test results have not been processed.\n";
-    }
+    $self->send_backed_up if $max_open_jobs <= keys %$jobs;
 
     return $jobs;
 }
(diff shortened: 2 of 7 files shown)


toddr:
This corrects the problems identified on #201  with too many file handles being open at once by the collector.

atoomic (approved):
tested and approved :-)

lib/Test2/Harness/Collector.pm
```diff
@@ -8,6 +8,9 @@ use Carp qw/croak/;
 
 use Test2::Harness::Collector::JobDir;
 
+# Max open files handles on a process applies. Usually this is 1024 so we can't have everything open at once when we're behind.
+use constant MAX_JOBS_TO_MONITOR => 200; # Allow for multiple handles per job to be open
```
exodist, line 12: I would prefer this be a runner option (See App::Yath::Options::Runner)

exodist: And I think the default should be undefined, with no limiting happening unless it is defined.

exodist (changes requested):
I want this configurable (you can add to your .yath.rc) with no behavior change by default.

This would also require documentation on when/why someone would use the option (your situation)

lib/Test2/Harness/Collector.pm
toddr: I very much disagree that default should be undefined. The real question is... if I have 2000 tests to parse, why should I open the file handles for all of them at once? Also what is the advantage to this being configurable? When would I want to bump this number and how would it be faster?

exodist: There are 2 main reasons:

1) The philosophy of Test and Specially Yath is to make EVERYTHING configurable. You can change the number of jobs, you can change if it uses TAP vs Stream, etc. This factors into it. People may decide they want to change the max number of open handles instead of limiting the collector. Their reasons may not be obvious to us.

2) When parsing STDOUT/STDERR and/or TAP without the Test2 event stream, timing data in the log is dependent on when the collector reads data, not when the data was written by the test. In your solution here a backed up collector will avoid reading new jobs until the backed-up ones are completely read, this directly goes against the design of the collector, which limits how much it reads from each job per iteration so that it can jump and read from all jobs. In your solution a handful of noisy jobs can offset timing data and display of other jobs. That is not necessarily wrong, but should be a choice, not a baked in design decision.

-- Additional Ideas (not quick, so probably not useful here, but worth mentioning anyway) --

If I were in this position I would work on making the tests that cause the collector to get backed-up less noisy, if their output is entirely important data (test results) then that is obviously not an option.

We might also consider making the collector more parallel (IE have 1 collector for every N job slots, also configurable) to prevent the backup, that would help with system/user filehandle limits a little bit by having less backups, but may not solve your problem. It would absolutely help with process fd limits.

toddr: As far as the collector parsing each job in parallel, I totally support this idea. It's just too much scope to fix right now.

toddr: I would suggest at at this point cPanel is the only group using the collector to this level. Anyone who lets the collector get to the point of there being more than 500 jobs to parse in one loop is going to crash. We picked 200 semi-arbitrarily but I think most people who hit this would be surprised. Which is why I suggest a default limit. I could go as high as 400 for the default but I think it's a mistake not to set it. I doubt very much anyone but us has hit this limit.

exodist:
I know ZipRecruiter used a -j56 when I was there on a super beefy machine. I do think we ran into this limit and simply bumped the max open filehandles limit.

ok, I am fine with a default, since it is a number of jobs we can stick with 200. But I want it configurable.

Also, I wonder if we can catch this exception and have it handle it gracefully... or if nothing else catch it, report that the user should probably lower this config options value, then die. That does not need to be part of this PR though if it is a lot of work.

toddr:
Concerning the `-j56`, I think some recent jitter in releases has somehow slowed down the collector. I was even wondering if we were accidentally not using `Cpanel::JSON::XS`. What we do know is that updating to the latest T2 has caused this to happen where it was not previously.

toddr:
We're totally ok with an option and documenting it. Updating the PR now.

exodist:
ok, yeah, add the option and docs, I will merge this, and then without a rush we can look into speeding up the collector (or fixing whatever slowed it down)

toddr:
@exodist: To add an option, am I going to have to create `lib/App/Yath/Options/Collector.pm` ?

exodist:
Hmm, yes, that may be necessary, any command that starts a collector (test and run currently) would need to include those options, so yeah they should be their own category/module. But that's the kind of flexibility this system was made to allow, so should not be hard to do. If you get stuck let me know, but I think everything is likely to get passed to all the right places already, so should not be a difficult change. Let me know if you get stuck.

lib/App/Yath/Command/collector.pm
```diff
@@ -16,6 +17,12 @@ use Test2::Harness::Run;
 use parent 'App::Yath::Command';
 use Test2::Harness::Util::HashBase;
 
+include_options(
+    'App::Yath::Options::Debug',
+    'App::Yath::Options::PreCommand',
```
toddr, line 22: I don't know if these other include options are appropriate here?

lib/App/Yath/Command/collector.pm
```diff
@@ -28,7 +35,7 @@ sub run {
 
     my $fh = isolate_stdout();
 
-    my $settings = Test2::Harness::Settings->new(File::Spec->catfile($dir, 'settings.json'));
+    my $settings = $self->settings;
```
toddr, line 38: I don't know the effect of removing settings.json here. Is this the right way to do it?

lib/Test2/Harness/Collector.pm
```diff
@@ -271,6 +281,11 @@ sub jobs {
         );
     }
 
+    # The collector didn't read in all the jobs because it'd run out of file handles. We need to let the stream know we're behind.
+    if(scalar %$jobs >= $max_jobs_to_process) {
+        print STDERR "WARN: The Yath Collector is running behind. More than $max_jobs_to_process test results have not been processed.\n";
```
toddr, line 286: This looks good in the output stream:

```
( PASSED )  job 273    t/test1.t
( PASSED )  job 361    t/test2.t
WARN: The Yath Collector is running behind. More than 300 test results have not been processed.
( PASSED )  job 267    t/test3.t
( PASSED )  job 279    t/test4.t
( PASSED )  job 366    t/test5.t
```
Is there a better way to pass this message to the formatter as a WARN? I see other random prints so I know it's not unheard of to do this?

lib/Test2/Harness/Util/Queue.pm
```diff
@@ -43,11 +43,12 @@ sub reset {
 
 sub poll {
```
toddr, line 44: Poll is taking a new arg now to pass through to poll_with_index().

exodist:
@toddr @atoomic I was going to explain how I needed things to change, but I had the time so I just added a commit instead. Now I need you guys to verify my changes still fix your problem. The changes you made would have fixed your problem, but the option was not actually usable and the collector would have been broken in subtle ways. If any of my changes need explanations please let me know.

If this does fix your issues I can merge and release.

lib/Test2/Harness/Collector.pm
exodist: See my additional commit which turns this into a proper events as you requested.

lib/App/Yath/Command/collector.pm
exodist: See my new commit

toddr:
I can't look till the morning but I am almost certain that 1000 will break it.￼ processing watch job requires multiple handles so 1000x2 is > 1024. It's really not important that the loop doesn't suck everything up since it'll get them on the next go around.

exodist:
Yeah, please do wait until you have time to fully read it, the 1000 you commented on is a different setting that has no effect on number of open handles, it is just a second option I chose to expose and the value is unchanged. I kept the 300 default that you had in your PR commit.

toddr:
Thanks for your help on this

lib/Test2/Harness/Collector.pm
```diff
@@ -214,16 +218,49 @@ sub process_tasks {
...
 
     return $jobs if $self->{+JOBS_DONE};
 
+    # Don't monitor more than 'max_open_jobs' or we might have too many open file handles and crash
+    # Max open files handles on a process applies. Usually this is 1024 so we
+    # can't have everything open at once when we're behind.
+    my $max_open_jobs = $self->settings->collector->max_open_jobs // 1024;
+    my $additional_jobs_to_parse = $max_open_jobs - keys %$jobs;
+    if($additional_jobs_to_parse <= 0) {
+        $self->send_backed_up;
```
toddr, line 257: This should never happen. If it does, it means an entire processing loop happened and no jobs were processed to completion. Given there are 300 jobs, that'd be bad or you passed a REALLY big number to -j.

lib/Test2/Harness/Collector.pm
```diff
@@ -271,6 +308,9 @@ sub jobs {
         );
     }
 
+    # The collector didn't read in all the jobs because it'd run out of file handles. We need to let the stream know we're behind.
+    $self->send_backed_up if $max_open_jobs <= keys %$jobs;
```
toddr, line 312: This will happen when backed up. It should be a different message than above. Ideally a small message to know that things are backed up here should be displayed every time it's backed up.

lib/Test2/Harness/Collector.pm
```diff
@@ -214,16 +218,49 @@ sub process_tasks {
...
+
+    # This is an unlikely code path. If we're here, it means the last loop couldn't process any results.
+    my $e = $self->_harness_event(0, undef, time, info => [{details => <<"    EOT", tag => "INTERNAL", debug => 1, important => 1}]);
+*** THIS IS NOT FATAL ***
+
+ * The collector has reached the maximum number of concurrent jobs to process.
+ * Testing will continue, but some tests may be running or even complete before they are rendered.
+ * All tests and events will eventually be displayed, and your final results will not be effected.
+
+Set a higher --max-open-jobs collector setting to prevent this problem in the
```
toddr, line 233: In our case, this doesn't help at all. The issue is that the number of events is too much to process and it is doing it sequentially. The collector just can't keep up. with even a single loop of processing. Incressing `--max-open-jobs` Will not make this better and will probably make it worse.

toddr (commented):
The changes look sane except for how and when it is messaging when it's backed up. We probably should be recommending they reduce the `-j` value, since setting it higher than CPU can cause a load which is pretty much the root cause of this problem. This also means that `-j` can never exceed 300 without significant complications.

toddr:
I've pushed a commit trying to straighten out the messaging. I'm going to try to test it now.

lib/Test2/Harness/Collector.pm
exodist: It happened in my testing when I set -j to 2 and set --max-open-jobs to 1. Not everyone will be using the defaults.

lib/Test2/Harness/Collector.pm
exodist: Take a look at my multiple-collector branch which has a better way to handle this. That branch is unstable, but I am going to bring some of the logic for determining when the issue happens back into this branch tonight when I have a chance. It should be better than what is here.

lib/Test2/Harness/Collector.pm
```diff
@@ -214,16 +218,49 @@ sub process_tasks {
...
+    my $e = $self->_harness_event(0, undef, time, info => [{details => <<"    EOT", tag => "INTERNAL", debug => 1, important => 1}]);
+*** THIS IS NOT FATAL ***
+
+ * The collector has reached the maximum number of concurrent jobs to process.
+ * Testing will continue, but some tests may be running or even complete before they are rendered.
+ * All tests and events will eventually be displayed, and your final results will not be effected.
+
+You may have set '-jX' too high. This can cause the system load to slow down the processing of
+results. It is also possible to set a higher '--max-open-jobs' collector setting, but be advised
+that could result in the run crashing due too many open filehandles.
```
exodist, line 235: maybe s/due/with/ ?

lib/Test2/Harness/Collector.pm
```diff
@@ -214,16 +218,49 @@ sub process_tasks {
...
+
+This message will only be shown once.
```
atoomic, line 237: what makes that message show once?

lib/Test2/Harness/Collector.pm
```diff
@@ -271,6 +308,12 @@ sub jobs {
         );
     }
 
+    # The collector didn't read in all the jobs because it'd run out of file handles. We need to let the harness output know we're behind.
+    if( $max_open_jobs <= scalar keys %$jobs ) {
+        my $msg = "The Yath Collector is running behind. More than $max_open_jobs test results have not been processed.";
+        $self->send_backed_up or $self->{+ACTION}->( $self->_harness_event(0, undef, time, info => [{details => $msg, tag => "INTERNAL", debug => 1, important => 1}]) );
```
atoomic, line 314: that message should not be displayed once

lib/Test2/Harness/Collector.pm
toddr: `$self->{+BACKED_UP}++;`

lib/Test2/Harness/Collector.pm
toddr: Right. That's what is happening.

lib/Test2/Harness/Collector.pm
toddr: yep typo. force push coming.

toddr: Fixed.

lib/Test2/Harness/Collector.pm
toddr: Sure but I would argue it's a bug to set --max-open-jobs  to less than -j and maybe we should warn on that?

lib/Test2/Harness/Collector.pm
toddr: Yep! thank you!

lib/App/Yath/Command/collector.pm
toddr: yep.

lib/App/Yath/Command/collector.pm
toddr: Thanks!

toddr:
This is currently failing make test but not CI

toddr:
We're testing this right now on our systems. Surprisingly even with `-j$nproc`, the collector runs behind pretty much the whole run until it gets to the end where single threaded tests run.

exodist:
This has been merged

PR #209, opened 2020-12-09
Two process management fixes

andfarm:
1. The logic for `no_set_pgrp` in the `start` command was reversed, causing it to not be used in the exact situation where it'd be most useful.

2. Avoid `close(STDIN)`. This wasn't causing any _obvious_ problems, but it did mean that some subprocesses could get created with `STDIN` pointing to the `yath` executable or to other unexpected files.

### DIFF
--- a/lib/Test2/Harness/Util/IPC.pm
+++ b/lib/Test2/Harness/Util/IPC.pm
@@ -105,7 +105,8 @@ sub _run_cmd_fork {
 
     swap_io(\*STDERR, $stderr, $die) if $stderr;
     swap_io(\*STDOUT, $stdout, $die) if $stdout;
-    $stdin ? swap_io(\*STDIN,  $stdin,  $die) : close(STDIN);
+    swap_io(\*STDIN,  $stdin,  $die) if $stdin;
+    open(STDIN, "<", "/dev/null") if !$stdin;
 
     @$cmd = map { ref($_) eq 'CODE' ? $_->() : $_ } @$cmd;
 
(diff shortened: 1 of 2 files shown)


lib/Test2/Harness/Util/IPC.pm
```diff
@@ -105,7 +105,8 @@ sub _run_cmd_fork {
 
     swap_io(\*STDERR, $stderr, $die) if $stderr;
     swap_io(\*STDOUT, $stdout, $die) if $stdout;
-    $stdin ? swap_io(\*STDIN,  $stdin,  $die) : close(STDIN);
```
exodist, line 108: I close STDIN for 2 reasons:
1) If I do not then the test may get keystrokes intended for something else. OR a test may sit there forever waiting for input and nobody types anything, best to close STDIN to make a `<STDIN>` fail. Though it appears that does not work?
2) Not closing STDIN in the test sometimes results in a borked terminal post-tests. I have not been able to figure out when/why it happens, but closing STDIN for test processes made it stop.

So, instead of leaving it open, lets figure out a better solution. Maybe always redirect STDIN to read from an empty file?

andfarm: > Maybe always redirect STDIN to read from an empty file?

That's precisely what I'm doing in L109. `open STDIN, "<", "/dev/null"` closes the old STDIN, then reopens `/dev/null` in its place (which will always return EOF on read).

haarg: IIRC on some systems, if you leave any of STDIN, STDOUT, or STDERR closed, opening new files may reuse those descriptors.  Attempts to read from STDIN could try to read from a random file opened in the future.

PR #212, opened 2021-02-12
Fix includes for Perl invocation

ylavoie:
Makes sure to fix PERL5LIB with includes only when Perl is used


exodist:
Can you explain what problem you were having/trying to fix? Maybe we can figure out an alternate solution.

PR #219, opened 2021-04-21
Replace last use of "base" with "parent"

JRaspass:
We already depend on parent, but bump it a little to ensure it has support for "-norequire".

### DIFF
--- a/lib/Test2/Formatter/Stream.pm
+++ b/lib/Test2/Formatter/Stream.pm
@@ -16,7 +16,7 @@ use Test2::Harness::Util qw/hub_truth apply_encoding/;
 
 use Test2::Util qw/get_tid ipc_separator/;
 
-use base qw/Test2::Formatter/;
+use parent qw/Test2::Formatter/;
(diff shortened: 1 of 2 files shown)


exodist:
Please note that the HashBase.pm and HashBase.t files are generated files, changes to them such as this will get overridden next time HashBase is regenerated. 
Hashbase is generated using https://metacpan.org/release/Object-HashBase
This module is also used to generate the hashbase used in Test-Simple, which cannot be switched to parent because it is such a fundamental dependency (It would cause a dep cycle)

I am fine with the non hashbase portions of this PR, so if you remove the HashBase changes (and leave base in the dep list) I can merge it.

If you REALLY want to remove 'base' from this codebase then you will first need to send a PR to the Object::HashBase repo (which is also mine) that adds a flag to the generator to replace 'base' with 'parent' in the generated results, then run that updated generator in Test2-Harness to make the changes.

Thanks for the PR and your effort on this, I do appreciate it, and I do hope that if the change matters to you that the tiny bit of necessary legwork is not going to stop you. The main issue here is that Test-Simple cannot break or have dep cycles, and this change requires making changes that could effect it. But I am happy to review as many PR's as you are willing to submit, and I will give you pointers as needed to make it happen.

JRaspass:
Ah gotcha, didn't realise those files came from another dist, I'm paired this commit down to the non Object::HashBase stuff for now.

Issue #112, opened 2019-08-22
Win32::Console::ANSI suggests prereq does not work


haarg:
The `Makefile.PL` in this dist attempts to add `Win32::Console::ANSI` as a suggested prerequisite on Win32.  It does this by adding it to `META_MERGE`.  This does not work.  If a `META.json` or `MEYA.yml` file exists, `META_ADD` and `META_MERGE` will be ignored.

It isn't possible to add suggests or recommends prerequisites dynamically using `ExtUtils::MakeMaker`, unless you monkey patch its internals.

exodist:
When did this behavior break? it worked fine when I was first directed to do it this way by the toolchain/perl-qa group.

haarg:
It's always been that way as far as I know.

dkechag:
It was also suggested to me on PerlMonks to use `META_MERGE` dynamically in order to have `requires` for cpantesters vs `recommends` otherwise as I wanted it, and it seems to work fine (on `Test2::Aggregate`) on all platforms.

exodist:
Currently yath does not even work/install on windows, so I am not going to bother addressing this unless yath on windows is eventually fixed and supported. Officially windows is not supported right now.

dkechag:
Yeah, while browsing to see how the refactor is going, I noticed you marked this a s a bug recently, when it does not even look like the original description is right about `ExtUtils::MakeMaker`.

haarg:
> It was also suggested to me on PerlMonks to use `META_MERGE` dynamically in order to have `requires` for cpantesters vs `recommends` otherwise as I wanted it, and it seems to work fine (on `Test2::Aggregate`) on all platforms.

How are you testing this?  Because I can assure you that my description is correct.

dkechag:
> How are you testing this? Because I can assure you that my description is correct.

Huh, looking into this, you may be right. I used this [Makefile.PL](https://metacpan.org/source/DKECHAG/Test2-Aggregate-0.12/Makefile.PL), to force cpantesters and cpancover to install extra packages for extended testing and it seemed to work. However, looking into it I think it may have worked because I did a development version where I had hard-set `requires` and it got the servers I was looking updated with the extra packages, so my subsequent versions seemed to work? I did not question it further since it seemed to be standard advice...

PR #232, opened 2021-09-28
Fix runtime-suggests line so it works

karenetheridge:
It is not possible, given the current features of ExtUtils::MakeMaker, to have dynamic prerequisites using the recommends, suggests or conflicts types. (This is because these get added via the META_ADD or META_MERGE Makefile arguments, and these are ignored for the generation of MYMETA.json.)


haarg:
Fixes #112

Issue #245, opened 2022-02-21
--color doesn't color everything


JRaspass:
Seems to only affect the final result line, not each job line
![Screenshot from 2022-02-21 10-10-16](https://user-images.githubusercontent.com/184356/154934431-c18dd7cc-0119-4845-8f4f-fd0fc40f0522.png)

PR #246, opened 2022-02-21
Color everything under --color

JRaspass:
Not sure I fully grasp the intended logic here but this appears to be the minimum fix for #245 without regressing use without the flag or use with the `--no-color` flag. The logic here could almost certainly be simplified, the nested testing of `$use_color` at least but I feel it's all more complex than it ought to be. Suggestions welcome.


JRaspass:
Also is it worth using `//=` for a lot of these `unless defined ...`, I see Test2::Harness depends on 5.10, I don't know if that was always the case.

PR #264, opened 2023-02-25
WIP Allow running a tainted runner

### LINKED ISSUE #262: Preload is useless with tainting
Given the following stripped down example:

`lib/MyExpensive.pm`:
```perl
package MyExpensive;

sleep 1;

1;
```
`lib/MyPreload.pm`:
```perl
package MyPreload;

use Test2::Harness::Runner::Preload;

stage DEFAULT => sub { default; preload 'MyExpensive' };

1;
```
`t/foo.t`:
```perl
use Test2::V0 -target => 'MyExpensive';

ok $ENV{T2_HARNESS_PRELOAD};

done_testing;
```

Running the tests with preload shows everything passes and the extra second isn't attributed to the test's "startup" cost:
```
$ yath test -TP MyPreload
( PASSED )  job  1    t/foo.t
(  TIME  )  job  1    Startup: 0.04258s | Events: 0.00018s | Cleanup: 0.00407s | Total: 0.04683s
```

However if we add a tainting shebang to the test file:
```perl
#!perl -T

use Test2::V0 -target => 'MyExpensive';

ok $ENV{T2_HARNESS_PRELOAD};

done_testing;
```

Then the test fails and takes longer:
```
$ yath test -TP MyPreload
[  FAIL  ]  job  1  + <UNNAMED ASSERTION>
(  DIAG  )  job  1    Failed test at t/foo.t line 5.
(  DIAG  )  job  1    Seeded srand with seed '20230215' from local date.
( FAILED )  job  1    t/foo.t
(  TIME  )  job  1    Startup: 1.08933s | Events: 0.00030s | Cleanup: 0.00322s | Total: 1.09285s
```

Now imagine that you have hundreds of tests, all with tainting enabled and lots of expensive libraries that take seconds to load, suddenly you're in my situation where $WORK's codebase takes tens of minutes to test :-(

I thought this was a regression since I'm sure this used to work.

I appreciate this is hard to make work as you can only adjust the taint mode at perl startup time and the worker you're forking doesn't have it enabled so can we either have multiple workers based on the set of interpreter flags we see in test files? Or an ugly hack to run all of yath under taint worker and all.

JRaspass:
**STILL WIP. DO NOT MERGE**

This allows tainted tests to benefit from the preload feature, fixes #262.

### DIFF
--- a/lib/App/Yath.pm
+++ b/lib/App/Yath.pm
@@ -17,7 +17,7 @@ use Test2::Harness::Util::HashBase qw{
 use Time::HiRes qw/time/;
 
 use App::Yath::Util qw/find_pfile/;
-use Test2::Harness::Util qw/find_libraries clean_path/;
+use Test2::Harness::Util qw/find_libraries clean_path mod2file/;
 use App::Yath::Options();
 use Scalar::Util qw/blessed/;
 
@@ -286,7 +286,7 @@ sub load_command {
     my ($cmd_name, %params) = @_;
 
     my $cmd_class = "App::Yath::Command::$cmd_name";
-    my $cmd_file  = "App/Yath/Command/$cmd_name.pm";
+    my $cmd_file  = mod2file($cmd_class);
 
     return $cmd_class if eval { require $cmd_file; 1 };
     my $error = $@ || 'unknown error';
--- a/lib/App/Yath/Command/test.pm
+++ b/lib/App/Yath/Command/test.pm
@@ -874,20 +874,26 @@ sub start_runner {
 
     my $settings = $self->settings;
     my $dir = $settings->workspace->workdir;
+    my @cmd = $^X;
+    my %env;
 
-    my @prof;
     if ($settings->runner->nytprof) {
-        push @prof => '-d:NYTProf';
+        push @cmd => '-d:NYTProf';
+        $env{NYTPROF} = 'start=no:addpid=1';
+    }
+
+    if ($settings->runner->taint) {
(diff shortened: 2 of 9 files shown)


exodist:
I am not bothered by the untaint, this proof of concept looks good. It will need a t/integration test that verifies it, as well as the cleanup and polish you already mentioned needing.

The one thing I would change about untaint() is that I would make it so that at load time it checks if taint mode is enabled, if it is then define sub untaint as you have, but if taint mode is not enabled just return the argument and skip the regex, as a minor performance enhancement when taint is not used.

The test is essential becaus eI will almost certainly break this feature without a test to save me from that. The test should skip on platforms/perls where taint mode is not available.

JRaspass:
Was this PR closed because the upstream branch was deleted or because you no longer want this change @exodist ?

exodist:
I was closed automatically when I deleted the wrong repo's master. I quickly restored master, but did not realize github auto-closed the PR.  The 're-open' button is grayed out, probably due to how long it was closed. Feel free to re-submit the PR. I have not had time, but I did not intentionally close it.

exodist:
ah, after my comment I can now reopen it, doing so.

exodist:
hmm, nevermind, the reopen did not work.

exodist:
Ok, I need to stop looking at PR's 5 minutes after waking up. This is not the repo that had master accidentally deleted, it is the one it was intentionally renamed to 1.0. I did not realize the name change would break PRs. Re-open against the 1.0 branch please.

PR #270, opened 2023-06-14
Update GitHub CI workflow

atoomic:
* Remove macOS and windows CI workflow

- Windows CI was already disabled.
- macOS workflow brings low value and use too much GitHub resources.
Let's focus in linux.

* Rename linux workflow to testsuite
* Add more Perl versions to test
* Bump actions/checkout
* Add a simple pre run before triggering the matrix


atoomic:
note: looks like github action got some issues, this is not related to that PR
view https://github.com/orgs/community/discussions/26693#discussioncomment-6177700
https://www.githubstatus.com is clean

PR #272, opened 2023-06-15
Disable T2_NO_FORK for CI

atoomic:
The test t/integration/reload_syntax_error.t fails under CI. 
Disable it for now.

### DIFF
--- a/.github/workflows/testsuite.yml
+++ b/.github/workflows/testsuite.yml
@@ -1,4 +1,4 @@
-name: linux
+name: testsuite
 
 on:
   push:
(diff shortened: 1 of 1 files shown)


atoomic:
and we have a winner :-) not sure why the test `t/integration/reload_syntax_error.t` fails for the Perl Tester images, whereas it works fine a basic ubuntu container...

atoomic:
note: this is not something new and was happening before the CI workflow refactor
view for example: https://github.com/Test-More/Test2-Harness/actions/runs/5267485952/jobs/9522930500
we should merge this case and if we want to investigate the issue with the test we can do it as a different case (or not)

PR #275, opened 2024-04-09
Pull the JUnit renderer into Yath for 2.0

exodist:
@toddr will you please review this? We discussed moving JUnit into Test2-Harness. I believe I already have your go-ahead, but I would like it formalized here with a PR approval from you.

This will not actually go to cpan for a few weeks (except maybe as early trials soon) as 2.0 is still under development. I am hoping to be able to finalize it or at least get close at the next PTS.


toddr:
I've not had a chance to look but yes I am ok with you taking these into T2::H. Just so you know, we're going to have to transfer PAUSE ownership. @atoomic will also be at toolchain so you can coordinate anything with him if you need to. You might ask PAUSE admins how you transfer a package INTO another bundle. I don't actually know how that works from a perms standpoint.

Issue #291, opened 2025-10-16
2.0 Branch PG upload fails on NULs, leaves the run in an indeterminate state in the UI


troglodyne:
OK, so for a pretty simple test like so:
`~/NUL-emitter.t`
```
use strict;
use warnings;

use Test2::Bundle::Extended;

isnt( "NUL\0", "NUL", "NUL\0 ain't NUL but it is ALSO \0" );

done_testing();
```
You'll get this when ran by itself, which is indicative of the fact that PGs past 9.4 can't handle NULs (ignore our DB config module, it's mostly glue code, important thing is we're using the PG upload route in code):
```
PERL5LIB=t/lib yath test --renderer DB --publish-mode complete --cover-files --project 'YATH_NUL_STALL_TEST' --db-config Test2::Harness::UI::Config::Cpanel ~/NUL-emitter.t
PERL_HASH_SEED not set, setting to '20251016' for more reproducible results.
No persistent runner, defaulting to the 'test' command.
System::Info is installed, setting job count to 4 (Half the cores on this system)
( PASSED )  job 1   ../../../root/NUL-emitter.t
[P:1|F:0|R:0|T:0] Events: 12
DBIx::Class::Storage::DBI::_dbh_execute_for_fetch(): execute_for_fetch() aborted with 'ERROR:  unsupported Unicode escape sequence
DETAIL:  \u0000 cannot be converted to text.
CONTEXT:  JSON data, line 1: ...b_try":0},"assert":{"pass":1,"details":"NUL\u0000...
unnamed portal parameter $5 = '...'' at populate slice:
{
  causes_fail => 0,
  event_idx => 7,
  event_sdx => 1,
  event_uuid => "0199EE09-323A-7D3B-8D62-8B6D89217EF9",
[...]

PR #292, opened 2025-10-22
Fix issue #291 - NULs aren't tolerated in PG

troglodyne:
Sadly, there are a *lot* of places where this "facet" data gets constructed, and not all of it goes through the same routes.
Since I couldn't actually find one place to scrub all this, I just wound up taking an experiential approach, finding tests that failed, poking at places that "looked right" and fixing things as the smoke went.

That said this pretty comprehensively solves the problem without having to go to heroic lengths. No idea how we want to go about unit testing this though.


troglodyne:
Wound up needing this in Stream.pm (scrubber sub):
```
return unless $INC{'Test2/Plugin/UTF8.pm'};
```
Presumably this is in fact a bug with the UTF8 plugin?

PR #294, opened 2025-10-22
Fix #293

### LINKED ISSUE #293: 2.0 Branch: Workdir in scope of Collector.pm is already a File::Temp dir, we don't need to clean up subsequent tempdirs
Noticed warnings caught by `Test::NoWarnings`/`FailWarnings`, etc. in the scope of most tests I was running where the parent pid created a tempdir in `App::Yath::Tester` that then got used as the `$workdir` in the scope of `Test2::Harness::Collector`. Sadly since that's already tracked for cleanup via the `UNLINK => 1` opt over in `Tester.pm`, you can hit a TOCTU within `File::Temp` that you'd never normally see (because you aren't generally expected to use `File::Temp` in this manner). Basically you hit `EACCESS` because the dir with the file which just passed a `-f` check got deleted right out from under you in-between that and the `unlink` call in `File::Temp`'s `END` block, as both that PID and the parent pid go out of scope at the same time.

Fix is to just not supply `UNLINK => 1` in `Collector.pm`.

troglodyne:
Don't unlink during END for your tempfile when a tempdir is already set to cleanup in END and passed in as the workdir.

Possibly I could be more intelligent about this and maybe check if workdir is *not* matching `^yath-` (set by App::Yath::Tester), but thus far I don't yet see a context where this wouldn't be true, so it just doesn' seem worth checking for.

### DIFF
--- a/lib/Test2/Harness/Collector.pm
+++ b/lib/Test2/Harness/Collector.pm
@@ -543,7 +543,7 @@ sub setup_child_input {
     }
     else {
         my $input = $ts->input // "";
-        my ($fh, $file) = tempfile("input-$$-XXXX", TMPDIR => 1, UNLINK => 1);
+        my ($fh, $file) = tempfile("input-$$-XXXX", TMPDIR => 1, UNLINK => 0);
         print $fh $input;
         close($fh);
         open($fh, '<', $file) or die "Could not open '$file' for reading: $!";


troglodyne:
Leaving some notes here on things I've noticed thus far when trying to add a test for this but I'm getting tired:
I need to find a way to force `Collector.pm`'s `setup_child_input` to have `+TEST_SETTINGS` with an `input` instead of `input_file`, as that's the route where this bug triggers. I'm guessing this involves forcing the `App::Yath::Tester::yath` invocation to have args that force usage of that so that we need to setup the `input-$$-XXXX` file there.

PR #295, opened 2025-11-16
Fix #291 - NULL characters cannot be stored in postgresql

### LINKED ISSUE #291: 2.0 Branch PG upload fails on NULs, leaves the run in an indeterminate state in the UI

exodist:
@troglodyne I would love your feedback on this fix for the PG null issue.

I tested it with `perl -Ilib scripts/yath -D test --publish-mode complete -v t/null.t --renderer=Server --ephemeral=PostgreSQL` and all seems good. Altering the commit to avoid escapes on PG reintroduces the problem.

This fixes things for PostgreSQL by adding an extra escape to all null characters just before decoding the json when using the run processor and a PostgreSQL database. This does not apply to other databases, but could if any end up with similar behavior.

I chose to escape them instead of stripping them as that makes it more obvious when someone looks at it. I actually suspect we may want to ALWAYS do this so that the null characters are always visible... But not yet.

This will find any \0 or \u0000 that has no escapes (IE the string literal is '\\0' or '\\u0000') And adds an escape. If there is already another escape in front of it then it is not modified, this avoids double-escaping any already escaped, and also will avoid adding escapes to nested json structures.

### DIFF
--- a/lib/App/Yath/Schema.pm
+++ b/lib/App/Yath/Schema.pm
@@ -50,6 +50,11 @@ sub is_mariadb {
     return 0;
 }
 
+sub can_store_null_character {
+    return 0 if is_postgresql();
+    return 1;
+}
+
 sub format_uuid_for_db {
     my $class = shift;
     my ($uuid) = @_;
--- a/lib/App/Yath/Schema/RunProcessor.pm
+++ b/lib/App/Yath/Schema/RunProcessor.pm
@@ -3,6 +3,8 @@ use strict;
 use warnings;
 use utf8;
 
+use feature qw/state/;
+
 our $VERSION = '2.000007';
 
 use DateTime;
@@ -21,7 +23,13 @@ use App::Yath::Schema::Config;
 
 use App::Yath::Schema::Util qw/format_duration is_invalid_subtest_name schema_config_from_settings format_uuid_for_db/;
 use Test2::Util::UUID qw/gen_uuid/;
-use Test2::Harness::Util::JSON qw/encode_json decode_json/;
+
+use Test2::Harness::Util::JSON qw/encode_json decode_json_no_null decode_json/;
+
+sub do_decode_json {
+    state $DECODER //= App::Yath::Schema::can_store_null_character() ? \&decode_json : \&decode_json_no_null;
+    return $DECODER->(@_);
+}
 
 use App::Yath::Schema::ImportModes qw{
     %MODES
@@ -205,7 +213,7 @@ sub _process_first_line {
         $dbh->{mysql_auto_reconnect} = 1 if $App::Yath::Schema::LOADED =~ m/(mysql|percona|maraidb)/i;
     }
 
-    my $e = decode_json(scalar $line);
+    my $e = do_decode_json(scalar $line);
     my $f = $e->{facet_data};
 
     my $self;
@@ -283,7 +291,7 @@ sub _process_first_line {
             my ($line, $idx) = @_;
 
             return if eval {
-                my $e = decode_json($line);
+                my $e = do_decode_json($line);
                 $self->process_event($e, undef, $idx);
                 1;
             };
--- a/lib/Test2/Harness/Util/JSON.pm
+++ b/lib/Test2/Harness/Util/JSON.pm
@@ -22,6 +22,7 @@ our @EXPORT_OK = qw{
 
     encode_json_file
     decode_json_file
+    decode_json_no_null
 };
 
 my $json   = Cpanel::JSON::XS->new->utf8(1)->convert_blessed(1)->allow_nonref(1);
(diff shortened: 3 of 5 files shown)


troglodyne:
Much cleaner fix. Thanks

PR #390, opened 2026-04-23
Implement CLI options

troglodyne:
Mostly just wiring up a few things, uses "old" 2.0 branch code where it made sense to do so.

Stubs out several things relied on by other modules for now so as to prevent explosions on loading modules.

Changes the behavior of command module loading to die when the role we want is not implemented by the command in question

Does not carry forward the init command.

Also adds App::Yath::Script to the cpanfile for what should be obvious reasons.


troglodyne:
Probably need to add tests for this but it's at least good enough to say whether I went in the wrong direction at least

lib/App/Yath2/IPC.pm
```diff
@@ -0,0 +1,7 @@
+package App::Yath2::IPC;
```
exodist: We do not want this module coming back, kill it.

lib/Test2/Harness2/Collector/Auditor/Run.pm
```diff
@@ -0,0 +1,6 @@
+package Test2::Harness2::Collector::Auditor::Run;
```
exodist: Not coming back, delete.

lib/Test2/Harness2/IPC/Protocol.pm
```diff
@@ -0,0 +1,6 @@
+package Test2::Harness2::IPC::Protocol;
```
exodist: Gone for good, delete

lib/Test2/Harness2/Collector/IOParser.pm
```diff
@@ -0,0 +1,6 @@
+package Test2::Harness2::Collector::IOParser;
```
exodist: Replaced by lib/Test2/Harness2/Collector/Parser/IOParser.pm

lib/Test2/Harness2/IPC/Util.pm
```diff
@@ -0,0 +1,6 @@
+package Test2::Harness2::IPC::Util;
```
exodist: Replaced by lib/Test2/Harness2/Util/IPC.pm

lib/Test2/Harness2/Util/LogFile.pm
```diff
@@ -0,0 +1,6 @@
+package Test2::Harness2::Util::LogFile;
```
exodist: Going to be replaced with a new App::Yath::LogArchive at some point

lib/Test2/Harness2/Client.pm
```diff
@@ -0,0 +1,6 @@
+package Test2::Harness2::Client;
```
exodist: Gone for good.

lib/Test2/Harness2/Instance.pm
```diff
@@ -0,0 +1,7 @@
+package Test2::Harness2::Instance;
```
exodist: Gone for good

lib/Test2/Harness2/TestSettings.pm
```diff
@@ -0,0 +1,272 @@
+package Test2::Harness2::TestSettings;
```
exodist: Replaced by Getopt::Yath::Settings

lib/Test2/Harness2/Util.pm
```diff
@@ -26,12 +26,18 @@ our @EXPORT_OK = qw{
     open_file
     parse_exit
     read_file
+    render_status_data
```
exodist: This will move to App::Yath2::Util, if we decide to carry it forward.

troglodyne:
Addressed commentary, now looking at what tests if any I can pull over/what extra I need to test here.

PR #403, opened 2026-04-24
Add finder role, fix bugs relating to it

troglodyne:
See commits for more details.

### DIFF
--- a/lib/App/Yath2/Finder.pm
+++ b/lib/App/Yath2/Finder.pm
@@ -17,6 +17,9 @@ use Test2::Harness2::Util::File::JSONL;
 use Test2::Harness2::TestFile;
 use File::Spec;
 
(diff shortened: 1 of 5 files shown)


lib/App/Yath2/Finder.pm
```diff
@@ -588,7 +597,7 @@ sub find_project_files {
 
     my $settings = $self->settings;
     my $default_search = [@{$self->default_search}];
-    push @$default_search => @{$self->default_at_search} if $settings->check_group('run') && $settings->run->author_testing;
+    push @$default_search => @{$self->default_at_search} if $settings && $settings->check_group('run') && $settings->run->author_testing;
```
troglodyne, line 600: This check may not be strictly necessary, but still probably best to be paranoid

PR #404, opened 2026-04-24
Add Renderers and filters

troglodyne:
See commit message for more details


troglodyne:
Going to rework this a bit, the Manager should be named outputmanager and the filter and renderer should be two discrete components that are just chained together, mostly because we may be able to reuse filters in other renderers.

lib/App/Yath2/Renderer/Default.pm
```diff
@@ -6,14 +6,14 @@ our $VERSION = '2.000011';
...
-use Test2::Harness2::Util qw/hub_truth apply_encoding mod2file fqmod/;
-use Test2::Harness2::Util::JSON qw/encode_pretty_json/;
+use App::Yath2::Theme;
+use Test2::Harness2::Util qw/hub_truth apply_encoding/;
 
 use File::Spec();
 use IO::Handle;
 use Scalar::Util qw/blessed/;
 use Storable qw/dclone/;
-use Test2::Util qw/IS_WIN32 clone_io/;
```
troglodyne, line 16: Removed because of being unused

lib/App/Yath2/Renderer/Default.pm
```diff
@@ -363,61 +316,24 @@ sub update_active_disp {
...
     }
 
     return $out unless $should_show;
 
-    my $theme = $self->{+THEME};
+    my $theme    = $self->{+THEME};
     my $statline = join '|' => (
-        $self->_highlight($stats->{passed},  'P', $theme->get_term_color(state => 'passed')),
-        $self->_highlight($stats->{failed},  'F', $theme->get_term_color(state => 'failed')),
-        $self->_highlight($stats->{running}, 'R', $theme->get_term_color(state => 'running')),
```
troglodyne, line 401: This change possibly is not desirable, need to revisit this

lib/App/Yath2/ArtifactLayer.pm
```diff
@@ -0,0 +1,180 @@
...
+
+    my $run_id = $args{run_id} or croak "'run_id' is required";
+    my $om     = $self->{+OUTPUT_MANAGER};
+
+    $om->start;
+
+    my %seen_done;
+    my $final;
+
+    while (1) {
```
troglodyne: This in particular I think needs to go away eventually just due to how the ipc/run logic will get refactored

Basically is here now just to make tests pass in a hobbled way

lib/Test2/Harness2/Spawn.pm
```diff
@@ -75,6 +75,12 @@ sub run_results {
     return $self->_send_request('run_results', \%args);
 }
 
+sub run_progress {
```
troglodyne: Probably will go away eventually

ARCHITECTURE.md
```diff
@@ -1122,111 +1122,179 @@ service stdout/stderr
...
+
+The pipeline has three distinct stages:
+
+```
+ArtifactLayer
+    │  (dispatches every event it reads — no mode filtering here)
+    ▼
+OutputManager
+    │  (holds one pipeline per renderer)
+    ├─► [Filter chain A] ──► Renderer A  (e.g. Default TUI)
```
exodist, line 1141: Each filter chain should also be able to send to multiple formatters, for example if we have an xml logger and a yaml logger that both need the same filtering, they can both live at the end of the same chain.

PR #424, opened 2026-04-25
Yath/integration/help.t: port from reference/old2 + dev install policy

atoomic:
## What

First slice of the `reference/old2` integration-test port: revives the
`App::Yath2::Tester` subprocess harness, brings back
`t/Yath/integration/help.t` verbatim, and stabilizes it so all 20
inner assertions pass.

Touches:
- `t/Yath/integration/help.t` — verbatim copy from `reference/old2`,
  asserts on `yath help`, `yath help help`, and `yath help test` output.
- `lib/Test2/Plugin/Immiscible.pm` — verbatim port of the flock mutex
  the integration tests need (only existed in `reference/old2`).
- `lib/App/Yath2.pm` — one-char regex fix in `app_path` so it strips the
  `Yath2.pm` basename, not just `Yath.pm` (V1 leftover).
- `lib/App/Yath2/Command/test.pm` — restore `summary` text and wire all
  11 expected option groups via `include_options(...)`.
- `author/install-yath-script` + CI `ubuntu-script-branch` job — SHA-pinned
  install of `App::Yath::Script` from `origin/script` so PRs that touch
  the (separately-distributed) wrapper get smoke-tested before release.
- `.gitignore` (`/local/`), `CLAUDE.md` (one-line testing note).

## How

1. **Tester wiring.** `App::Yath2::Tester` already shipped here from old2
   but `app_path()` returned the `.pm` file path instead of `lib/`,
   which fed a malformed `-I`/`-D` to the spawned yath and hung it.
   Regex `App\S+Yath\.pm$` → `App\S+Yath\d*\.pm$`.
2. **Immiscible.** Plain `cp` from `reference/old2/lib/Test2/Plugin/`.
3. **help.t.** Plain `cp` from `reference/old2/t/Yath/integration/`.
4. **`Command::test` regressions.** The verbatim port surfaced two:
   - `summary` had drifted to `'Run a list of test files'`, so
     `qr{test.+Run tests}m` no longer matched. Restored to old2's
     `'Run tests with a clean temporary runner'`.
   - Only `Workspace` + `Yath` option groups were wired; old2 had
     eleven. `include_options(...)` extended to the full list (Yath,
     Harness, Workspace, Finder, IPC, Renderer, Resource, Run,
     Runner, Scheduler, Term, Tests). All modules already exist in
     `lib/App/Yath2/Options/`.
5. **Install policy.** New `author/install-yath-script` does
   `git rev-parse origin/script`, compares against
   `local/.app-yath-script-sha`, and runs `cpm`/`cpanm` from a detached
   worktree only when the SHA changed. Idempotent re-runs are no-ops.
   New CI job `ubuntu-script-branch` runs the same suite as
   `Simple TestRun` but installs `App::Yath::Script` from
   `origin/script` first; existing CI jobs untouched and keep using the
   CPAN release.

## Why

- Need the integration-test framework working end-to-end before the
  remaining 34 tests in `reference/old2/t/Yath/integration/` can be
  ported.
- Want a cheap loop for testing the unreleased
  `App::Yath::Script` wrapper without cutting CPAN releases — the
  CI side validates `origin/script` on every PR; the dev side
  re-installs only on SHA change.
- The two `Command::test` fixes are the smallest possible patches
  that make the help-output contract stick. Wider work
  (`Command::test` actually consuming the new options at runtime,
  centralizing `Command::help`'s stale-load filtering) is
  out of scope and tracked as follow-ups.

## Verification

`perl -Ilib t/Yath/integration/help.t` exits 0; 20/20 inner
assertions pass:

- subtest 1 `yath help` — 5/5 (exit, usage, help/test/start listed)
- subtest 2 `yath help help` — 2/2
- subtest 3 `yath help test` — 13/13 (all 11 option-group labels)

### DIFF
--- a/lib/App/Yath2.pm
+++ b/lib/App/Yath2.pm
@@ -47,7 +47,7 @@ use Test2::Harness2::Util qw/find_libraries clean_path mod2file read_file/;
 use Test2::Harness2::Util::JSON qw/encode_pretty_json decode_json/;
 
 my $APP_PATH = __FILE__;
-$APP_PATH =~ s{App\S+Yath\.pm$}{}g;
+$APP_PATH =~ s{App\S+Yath\d*\.pm$}{}g;
 $APP_PATH = clean_path($APP_PATH);
 sub app_path { $APP_PATH }
 
--- a/lib/App/Yath2/Command/test.pm
+++ b/lib/App/Yath2/Command/test.pm
@@ -29,8 +29,18 @@ use App::Yath2::Renderer::Default();
 
 use Getopt::Yath;
 include_options(
-    'App::Yath2::Options::Workspace',
     'App::Yath2::Options::Yath',
+    'App::Yath2::Options::Harness',
+    'App::Yath2::Options::Workspace',
+    'App::Yath2::Options::Finder',
+    'App::Yath2::Options::IPC',
+    'App::Yath2::Options::Renderer',
+    'App::Yath2::Options::Resource',
+    'App::Yath2::Options::Run',
+    'App::Yath2::Options::Runner',
+    'App::Yath2::Options::Scheduler',
+    'App::Yath2::Options::Term',
+    'App::Yath2::Options::Tests',
 );
 
 use Role::Tiny::With;
--- a/t/Yath/integration/help.t
+++ b/t/Yath/integration/help.t
@@ -1,0 +1,62 @@
+# HARNESS-CONFLICTS YATH
+use Test2::V0;
+
+use File::Temp qw/tempdir/;
+use File::Spec;
+
+use lib 't/lib';
+use Test2::Harness2::Test::Yath qw/yath/;
+use App::Yath2::Util qw/find_yath/;
+
+yath(
+    command => 'help',
+    args    => [],
+    exit    => 0,
+    test    => sub {
+        my $out = shift;
+
+        like($out->{output}, qr{^Usage: .*yath}m, "Found usage statement");
+
+        # Sample some essential commands
+        like($out->{output}, qr{help.+Show the list of commands}m,         "'help' command is listed");
+        like($out->{output}, qr{test.+Run a list of test files}m,          "'test' command is listed");
+        like($out->{output}, qr{start.+Start a test runner}m, "'start' command is listed");
+    },
+);
+
+yath(
+    command => 'help',
+    args    => ['help'],
+    exit    => 0,
+    test    => sub {
+        my $out    = shift;
+        my $script = find_yath();
+
+        like($out->{output}, qr/Command selected: help/, "Showing help for the help command");
+    },
+);
+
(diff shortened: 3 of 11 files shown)


t/Yath/integration/help.t
```diff
@@ -0,0 +1,62 @@
+use Test2::V0;
+
+use File::Temp qw/tempdir/;
+use File::Spec;
+
+use App::Yath2::Tester qw/yath/;
+use App::Yath2::Util qw/find_yath/;
+
+use Test2::Plugin::Immiscible(sub { $ENV{TEST2_HARNESS_ACTIVE} ? 1 : 0 });
```
troglodyne: Perhaps better done via HARNESS-CONFLICTS comment?

lib/App/Yath2/Command/test.pm
```diff
@@ -29,16 +29,26 @@ use App::Yath2::Renderer::Default();
...
+    'App::Yath2::Options::Tests',
 );
 
 use Role::Tiny::With;
 with 'App::Yath2::Role::Command';
 
 sub args_include_tests { 1 }
 sub group              { 'test' }
-sub summary            { 'Run a list of test files' }
+sub summary            { 'Run tests with a clean temporary runner' }
```
troglodyne: Not sure why this needs updating?

lib/App/Yath2.pm
```diff
@@ -47,7 +47,7 @@ use Test2::Harness2::Util qw/find_libraries clean_path mod2file read_file/;
 use Test2::Harness2::Util::JSON qw/encode_pretty_json decode_json/;
 
 my $APP_PATH = __FILE__;
-$APP_PATH =~ s{App\S+Yath\.pm$}{}g;
+$APP_PATH =~ s{App\S+Yath\d*\.pm$}{}g;
```
troglodyne, line 50: I'm guessing this was why the test would hang previously once imiscible issue was resolved (not specifying the right yath)?

lib/Test2/Plugin/Immiscible.pm
troglodyne, line 1: This should probably have commentary/POD about why this is still needed/what use cases it still has if the HARNESS-CONFLICTS comment isn't sufficient to make this work right

lib/App/Yath2.pm
atoomic: yeap going to add a sanity check from unit test to confirm we are loading the expected local version

lib/Test2/Plugin/Immiscible.pm
atoomic: yeap going to remove it with some doc regarding it

t/Yath/integration/help.t
atoomic: fixed, not needed and switched to `# HARNESS-CONFLICTS YATH`

PR #426, opened 2026-04-25
t/Yath/integration: port the remaining 34 tests with skip_all gates

atoomic:
Bring back the rest of reference/old2/t/Yath/integration/ (34
tests + their fixture trees) so the integration surface is in-tree
and runnable today, even though most of it depends on features
the rewrite has not re-implemented yet.

For every port:

  - The opening lines follow the pattern established for help.t:
    `# HARNESS-CONFLICTS YATH` directive at the top, the t/lib
    wrapper instead of `use App::Yath2::Tester` directly, and no
    Test2::Plugin::Immiscible (the on-disk flock fallback was
    removed earlier in this branch).

  - Tests that depend on features not yet implemented in this
    branch are gated with:

        plan skip_all => "TODO: <reason>";
        __END__

    The original body is preserved untouched after `__END__` so
    re-enabling a test once its feature lands is just a matter of
    deleting the prelude.

Aggregate after the port (run via `perl -Ilib t/Yath/integration/*.t`
on a clean checkout):

  - PASS (real assertions):   help, test-w, verbose_env       (3)
  - AUTHOR-gated (preexisting): coverage4, reload_syntax_error,
                                signals                         (3)
  - HARNESS-ONLY (preexisting): slots_per_job{,2,3}             (3)
  - TODO skip_all (this commit): 26 tests

  Total: 35/35 exit 0.

The TODO reasons name the missing feature (e.g. "coverage
instrumentation (--cover-write/--cover-from) not implemented",
"persistent runner daemon not implemented", "speedtag command
output not aligned with current log format"), so the suite doubles
as a punch list for the rewrite.

AI_DOCS/2026-04-25-integration-test-inventory.md captures the
full status table by category and documents the
"delete-prelude-to-revisit" workflow.

Co-Authored-By: Claude <noreply@anthropic.com>


atoomic:
Note that all integration tests not available there are put back with a TODO / skip_all giving us a plan to re enable them one after the other without worrying about leaving some behind

PR #431, opened 2026-04-25
broken symlink and init

atoomic:
- **dist.ini: exclude broken-symlink fixture from dzil gather**
- **App::Yath2::Command::init: port from reference/old2**


dist.ini
```diff
@@ -45,6 +45,14 @@ exclude_filename = cpanfile
 root = t/
 prefix = t/
 include_dotfiles = 1
+; This is a deliberately-dangling symlink (-> nothing-there) used as a
```
exodist: Do not track this in git if the test creates it.

PR #434, opened 2026-04-25
log_dir.t: wire --log / -L and --log-dir on Command::test

atoomic:
t/Yath/integration/log_dir.t passes a tempdir via --log-dir, sets
-L to enable logging, and asserts that exactly one .jsonl file
ends up in the tempdir. Until now Command::test only had
--log-file / -F (commit b3ed09610), which forces the caller to
hand-craft a path; the test wants the runner to do the path
synthesis itself.

Add the missing piece on Command::test:

  --log-file / -F PATH  explicit path (already there)
  --log     / -L        Bool: turn logging on
  --log-dir DIR         Scalar: directory for the synthesized
                        filename when --log is on without --log-file

Resolution in run():
  1. --log-file wins outright
  2. else, if --log is on, synthesize
        File::Spec->catfile($log_dir // '.', strftime('%Y%m%d-%H%M%S').'.jsonl')
  3. else, no user log

`alt => ['log']` was dropped from the log_file option to free the
plain `--log` name for the new Bool. Existing callers that pass
--log=PATH would now hit the Bool option; any in-tree caller uses
-F or --log-file directly so this is a no-op for our suite (Tester's
`log => 1` arg passes -F /tmp/yathlog-XXX.jsonl, not --log).

log_dir.t flips from skip to pass with three inner ok:

  ok 1 - Exit Value Check
  ok 2 - Only 1 file present
  ok 3 - File is a jsonl file

dzil test stays green: 130 files / 799 tests / Result: PASS. The
other already-active integration tests (help, init,
nested_includes, test-w, verbose_env) still pass through the new
log resolution path because none of them set -L; the path stays
undef and the tee logic short-circuits exactly as before.

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/lib/App/Yath2/Command/test.pm
+++ b/lib/App/Yath2/Command/test.pm
@@ -12,9 +12,8 @@ use Object::HashBase qw{
     <plugins
 };
 
-use File::Spec();
 use File::Path qw/remove_tree/;
-use Carp qw/croak/;
+use File::Spec();
 use POSIX qw/strftime/;
 use Time::HiRes qw/sleep/;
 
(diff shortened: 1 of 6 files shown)


atoomic (commented):
need work: we want to match the new log file type

PR #437, opened 2026-04-25
Bring back more integration tests (part 1)

troglodyne:
Needed to first fix some things with the log location in the test lib.

Otherwise restored the following:

- replay.t: runs yath test + yath replay; asserts identical output
- failed.t: runs yath test + yath failed; asserts fail.tx listed
- times.t: runs yath test + yath times; asserts timing table headers
- speedtag.t: runs yath test + yath speedtag; asserts HARNESS-DURATION tags written

replay.t and speedtag.t are marked HARNESS-DURATION-SLOW/MODERATE since
they make two full yath invocations each.

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

### DIFF
--- a/t/Yath/integration/replay.t
+++ b/t/Yath/integration/replay.t
@@ -1,8 +1,5 @@
 # HARNESS-CONFLICTS YATH
-use Test2::V0;
-plan skip_all => "TODO: replay command not yet aligned with current log/streamer format";
-__END__
-
+# HARNESS-DURATION-SLOW
 use Test2::V0;
 
 use File::Temp qw/tempdir/;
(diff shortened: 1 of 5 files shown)


t/Yath/integration/replay.t
```diff
@@ -21,6 +18,7 @@ sub clean_output {
     my $out = shift;
     $out->{output} =~ s/^.*duration.*$//m;
     $out->{output} =~ s/^.*Wrote log file:.*$//m;
+    $out->{output} =~ s/^.*Wrote archive:.*$//m;
```
troglodyne, line 21: This will eventually need removal when we stop automatically writing the archive.

PR #439, opened 2026-04-25
restore a few more integration tests

troglodyne:
Basically needed to force -j 1


troglodyne:
Still working on fixing this

PR #441, opened 2026-04-25
feat(options): add --no-plugins to disable auto-plugin discovery

troglodyne:
## What

Add `--no-plugins` Bool option to `App::Yath2::Options::Yath`; wire it into the auto-plugin discovery loop in `App::Yath2::process_args`.

## Why

`stamps.t` uses `--no-plugins -pTestPlugin` to load only one specific plugin without ambient auto-discovered plugins interfering. Without this flag there is no way to isolate plugin loading in tests. Note: `stamps.t` remains blocked by a separate issue (`$log->poll()` JSONL API), but `--no-plugins` is independently useful for test isolation in any context.

`--log-dir` / `-L` were already implemented in `Options::Logging` and wired in `Command::test` (PR #434 merged it); `log_dir.t` already runs without a skip guard. No work needed there.

## How

- `Options::Yath.pm`: add `no_plugins => Bool` option with `alt => ['no-plugins']`, default 0
- `Yath2.pm` `process_args`: read `$settings->yath->no_plugins` (wrapped in eval for safety); in the `STATE_MODULES` loop, skip any `App::Yath2::Plugin` subclass when the flag is set. Renderers and resources are unaffected.
- Explicitly requested plugins (`-p Foo`) are processed separately in `run()` and are not affected.

## Testing

Unit test: pass `--no-plugins` to `yath test` alongside `-pSomePlugin`; confirm only the explicit plugin is active.  
Full integration: `stamps.t` will exercise this once the `$log->poll()` API issue is resolved.

### DIFF
--- a/lib/App/Yath2.pm
+++ b/lib/App/Yath2.pm
@@ -506,7 +506,7 @@ sub process_args {
     $self->{+ENV_VARS} = $self->{+STATE_ENV};
     $self->{+OPTION_STATE} = $state;
 
-    my $no_plugins = eval { $settings->yath->no_plugins } // 0;
+    my $no_plugins = eval { $settings->yath->no_plugin_scan } // 0;
     for my $module (keys %{$self->{+STATE_MODULES}}) {
         for my $set (['yath', 'plugins', 'App::Yath2::Plugin'], ['renderer', 'classes', 'App::Yath2::Renderer'], ['resource', 'classes', 'App::Yath2::Resource']) {
             my ($group, $field, $type) = @$set;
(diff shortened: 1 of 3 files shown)


troglodyne:
### CI note: Perl 5.24/5.28 failures on replay.t

The `Simple TestRun` (system Perl) passes. Two older Perl version jobs fail on `t/Yath/integration/replay.t` — the same interleave ordering issue that PR #443 addresses.

Our `clean_output` normalization (sort all consecutive `job N` lines) fixes the flap on newer Perls, but on 5.24/5.28 the output appears to have non-job lines interspersed between the job-status lines at different positions between live and replay runs, defeating the consecutive-group sort.

This is pre-existing — the test was behind `skip_all` precisely because it wasn't reliable. The fix for older Perls likely requires sorting ALL `job N` lines globally (ignoring non-job lines between them), rather than only consecutive groups. That's a follow-up; the primary feature (`--no-plugins`) is unaffected.

troglodyne:
Solved by other work actually :)

PR #442, opened 2026-04-26
speedtag.t: fix Command::speedtag's accessor

atoomic:
- **speedtag.t: fix Command::speedtag's accessor + duration fallback**
- **Streamer::Static: extract bundled zstd-dict to tempdir**
- **LogArchive::Directory: add dict_bytes for live logdirs**

### DIFF
--- a/lib/App/Yath2/LogArchive/Directory.pm
+++ b/lib/App/Yath2/LogArchive/Directory.pm
@@ -10,7 +10,8 @@ use Role::Tiny::With;
 use parent 'App::Yath2::LogArchive';
 use Object::HashBase qw/path format/;
 
-with 'App::Yath2::LogArchive::Role::Source';
+with 'App::Yath2::LogArchive::Role::Source',
+     'App::Yath2::LogArchive::Role::DiskDict';
 
 sub viable { 1 }
 
--- a/lib/App/Yath2/Streamer/Static.pm
+++ b/lib/App/Yath2/Streamer/Static.pm
@@ -113,8 +113,7 @@ sub _resolve_path {
 
     return undef unless $archive->has_file($rel);
 
-    my $tmpdir = $self->{+ARCHIVE_TMPDIR} //=
-        tempdir('yath-streamer-XXXXXX', TMPDIR => 1, CLEANUP => 1);
+    my $tmpdir = $self->{+ARCHIVE_TMPDIR} //= $self->_make_archive_tempdir($archive);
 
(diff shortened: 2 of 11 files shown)


lib/App/Yath2/LogArchive/Directory.pm
```diff
@@ -34,6 +36,18 @@ sub has_file {
     return -f $abs ? 1 : 0;
 }
 
+# Mirror TarZIdx's dict_bytes: when the on-disk logdir carries a
+# zstd-dict.bin at its root (the harness drops one there at run start
+# when DICT_PATH resolves), return its bytes so artifacts() and other
+# dict-aware readers can decompress dict-compressed snapshots.
+# Returns undef when the logdir is dict-less.
+sub dict_bytes {
```
troglodyne: Why mirror? does this instead need moving out into a tool lib for reuse?

atoomic: yeap fix moving on that

atoomic: moved to a custom role

lib/App/Yath2/Streamer/JSONL.pm
exodist, line 1: We should not be working with .jsonl final log artifacts anymore, so this streamer should be dropped.

lib/App/Yath2/Streamer/Static.pm
```diff
@@ -134,6 +133,37 @@ sub _resolve_path {
...
+# Role::Source dict_bytes contract returns dictionary bytes, drop
+# them at the tempdir root so per-logger log_reader()s -- which
+# discover the dict by walking up from each .zst snapshot's
+# directory looking for an adjacent zstd-dict.bin -- find it.
+# Without this any dict-compressed snapshot extracted from the
+# archive croaks with "Dictionary mismatch" when read back.
+#
+# The dict is binary; the open uses binmode explicitly because
+# Test2::Harness2::Util's file helpers do not.
+sub _make_archive_tempdir {
```
exodist, line 137: Is this just an over complicated way to copy the dictionary file from one location to another?

exodist:
Couple comments, JSONL log format is going away. There is a method where I am not sure what it is doing or why, but it might just be a complicated copy-file?

Only other nit is that I would like the git commit history consolidated to trim out dead-end changes and things that were added then removed or changed. Clean history.

PR #443, opened 2026-04-26
replay.t: stop flapping on per-job event interleave

atoomic:
## Summary

`t/Yath/integration/replay.t` is a flapper. CI hit it on run [#24952534922](https://github.com/Test-More/Test2-Harness/actions/runs/24952534922/job/73065200704) (linux Perl 5.26): attempt 1 failed, attempt 2 passed.

The test runs `yath test`, captures the live stdout, then runs `yath replay` on the saved log archive and asserts both outputs are byte-identical after `clean_output` normalisation. The live and replay renderers emit the same set of per-job lines but in different orders depending on how the failed job's diagnostics interleave with another job's `( PASSED )` status line:

- **Live** (timing-dependent): the failed job's `[ FAIL ]` / `( DIAG )` / `< REASON >` lines drain through the renderer pipeline before `pass.tx`'s status line arrives — `( FAILED ) fail.tx → diags → ( PASSED ) pass.tx`.
- **Replay**: status lines emit first, then all the per-test events trail at the end — `( FAILED ) fail.tx → ( PASSED ) pass.tx → diags`.

The previous `clean_output` only sorted **consecutive** `( PASSED )` / `( FAILED )` status lines. That fired in the replay case (where the two status lines ARE consecutive) but not in the live case (the `[ FAIL ]` / `( DIAG )` / `< REASON >` lines break the run), so the two normalisations ended up in different orders and the equality assertion flapped.

## Fix

Trying to attach diag lines back to "their" status block is fragile because the diags carry only `job N` (already normalised to a sentinel by `clean_output`) — there is no filename to disambiguate which status they belong to.

The robust normalisation is to canonicalise the order of every `job N`-prefixed line:

- Walk the cleaned output line by line.
- Buffer consecutive `job N` lines into a run.
- On hitting a non-job line, sort the run lexicographically and flush, then pass the non-job line through.

Non-job lines (the "The following jobs failed:" table, the "Yath Result Summary" block, etc.) keep their position so the file's overall narrative survives.

## Test plan

- [x] 20/20 back-to-back local runs of `yath -D test t/Yath/integration/replay.t` pass.
- [x] A simulated diff of the exact CI failure case (live with interleaved diags vs replay with status-first) normalises to identical strings under the new algorithm.
- [x] Watch the next CI run on this branch to confirm the flake is gone.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


troglodyne (approved):
Probably could have been done in less lines with a more clever map, but not worth the effort to golf it down, especially since it is a test.

May not be the right fix if you think the ordering is in fact important here @exodist

t/Yath/integration/replay.t
```diff
@@ -48,23 +48,30 @@ sub clean_output {
...
+    # is to canonicalise the order of every `job N`-prefixed line:
+    # collect consecutive runs of them, sort the run, flush. Non-job
+    # lines (the "The following jobs failed:" table, the "Yath
+    # Result Summary" block, etc.) pass through verbatim so the
+    # file's overall narrative stays intact.
     my @normalized;
-    my @status_group;
+    my @run;
     for my $line (@lines) {
-        if ($line =~ /^\(\s*(?:PASSED|FAILED)\s*\)/) {
```
atoomic, line 56: it does not really makes sense why we remove that regexp there, as we are going to do a sort later

t/Yath/integration/replay.t
```diff
@@ -48,23 +48,30 @@ sub clean_output {
         push @lines => $line;
     }
 
-    # Sort consecutive PASSED/FAILED job-status lines so that parallel
-    # completion order does not break the comparison.
+    # Live (`yath test`) and replay emit the same per-job lines but
+    # in different orders, depending on how the failed job's
```
atoomic: this is why we use a `sort` later

t/Yath/integration/replay.t
troglodyne: well yes and no, the order of `@normalized` depends on the order of `@run`. Operative changes do seem to be the changed regexp and changed behavior around that. Any non-notification line seems to drain `@run` now (whereas previously, non "pass/fail" events drained it too). 

The last sort on `@run` is kind of a catchall in case we run out of lines by the look of things. Thus my thinking this could be golfed down with a more clever map.

t/Yath/integration/replay/expected_output.txt
atoomic, line 1: is it a cruft?

atoomic: nope: looks like used later

atoomic: note: this is fragile if we decide to update the test format 
we will not notice the regression here

troglodyne: That's why I generated saved archive generated from 1.0 branch

If we change format of output and it doesn't break old replays then we probably haven't deduplicated enough code in the default renderer. Both go through the OutputManager so I don't think this should be the case.

If we suddenly replace [ PASSED ] with << Passed >> or whatever that should break due to us saving the "golden" expected text.

Are you saying that if we suddenly *introduce* new event types that these won't be covered? Yea sure I guess so, though if we drop event types via filters or something then the test will fail as the filters are chained together to the renderer via the OutputManager, so I doubt dropping event types we already expect here would result in anything other than a failure.

troglodyne: A way we could put a stop to this is a test that doesn't care about order but instead generates every possible event type then replays them. Then it's just iterating over lines populating a %seen hash then asserting on keys after the replay.

atoomic: We can still improve that replay test later as needed, for now it s kind of blocking other PRs
this is fine to use it that way I reverted cruft previously pushed there to restore your state

PR #452, opened 2026-08-06
Collector: sleep with exponential backoff when idle instead of busy-looping

dmitry-novozhilov:
The main collector loop only slept when $count was 0, but $count is incremented unconditionally for every active job (once for existing, once more for not being done yet). As a result the loop never slept while at least one test was running, polling job dir files as fast as possible and pinning a full CPU core for the duration of the run, regardless of concurrency settings.

Track actually-processed work (runner output lines, queue tasks, job events) separately from the liveness counter that guards loop exit. When an iteration processes nothing, usleep with exponential backoff starting at 1ms and doubling up to a 100ms cap; any activity resets the backoff to 1ms.

Note: the collector's wait_time attribute (default 0.02s, not exposed as a CLI option and never set from within the dist) is no longer read by this loop.

P.S.: claude used.

### DIFF
--- a/Changes
+++ b/Changes
@@ -2,6 +2,10 @@
 
     - Add [PruneCruft] to dist.ini so build artifacts (blib/,
       pm_to_blib, MYMETA.*) no longer leak into release tarballs.
+    - Fix collector busy-loop: sleep with exponential backoff (1ms..100ms)
+      when no events were processed, instead of never sleeping while any
+      job is active. Collector CPU usage drops from 100% of a core to ~0
+      during normal runs.
 
 1.000172  2026-04-28 21:15:51-07:00 America/Los_Angeles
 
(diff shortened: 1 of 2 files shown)


kfix:
FWIW I've tried this patch and it has fixed a long standing busy-looping yath-collector process in our CI's testing runs with no discernible impact on their logging.

exodist:
I will attempt to take a look at this soon, and assuming all is good merge and release, sorry it has taken so long, been very busy. I appreciate the AI use transparency, I have no objections to using AI to assist in coding for this project :-)

exodist:
Can you look at https://github.com/Test-More/Test2-Harness/pull/453 and see if your problem is still solved? I had an agent of my own do some corrections based on some multi-repo documentation I am training to help me do code reviews of PRs. If it does not still solve your problem that would be important to know.

PR #455, opened 2026-08-18
shared locks: like LOCK_SH for HARNESS-CONFLICTS's LOCK_EX

rjbs:
Right now if you have ten tests, and one of them can't run at the same time as the others, you end up putting HARNESS-CONFLICTS-WHATEVER at the top of all ten, and now they're strictly serialized.

With this branch you get HARNESS-SHARES-WHATEVER.  The nine friendly tests get that, and can run alongside each other, but the greedy test keeps HARNESS-CONFLICTS-WHATEVER and now won't run with any of the others.

If you declare both in one file, it's exclusive and warns.

Bonus: tests for Test2::Harness::Runner::State

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,5 +1,7 @@
 {{$NEXT}}
 
+    - Add the # HARNESS-SHARES-XXX directive, a shared claim on a name: any number of tests sharing XXX run at the same time, but none run alongside a # HARNESS-CONFLICTS-XXX test, which claims XXX exclusively.
+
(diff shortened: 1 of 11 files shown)


rjbs:
CI failures on older Perls are because the workflow uses cpm, which requires a more recent perl.  I tested on v5.14 manually and it's fine.  (I made an offer elsewhere to rewrite your test workflow, lmk.)

